PR #11, opened 2025-06-19
Simplify API for insight hub and also handle pagination

sazap10:
## Goal

Implement just the API endpoints we are using in the Insight Hub tools/resources. Also adds ability to paginate responses. 

## Design

Makes the API more readable and easier to maintain

## Changeset

Removes the current API implementation and replaces with a smaller API with endpoint just used in the MCP server

## Testing

Tested each tool and resource to ensure they still work


yousif-bugsnag (approved):
LGTM

PR #13, opened 2025-06-20
Configure Node CI

gingerbenw:
## Goal

Added a GitHub workflow to ensure TS code compiles and source files pass the linting rules when code is pushed

## Design

This PR uses ESLint which is the defacto code quality tool across JS repositories, and runs on GitHub actions to avoid the requirement for additional third party tooling

## Changeset

Installed and configured ESLint


gingerbenw:
@sazap10 Do you know if any of the files here are generated? we should probably ignore linting on generated files

sazap10:
> @sazap10 Do you know if any of the files here are generated? we should probably ignore linting on generated files

nothing should be generated right now

sazap10:
i get a bunch of errors locally, im surprised they dont show in CI

sazap10 (approved):

PR #15, opened 2025-06-20
OP-33707: Publish to NPM

JonathanParrilla:
## Goal

As per https://smartbear.atlassian.net/browse/OP-33707, we need to publish to NPM under the SmartBear account.

## Design

Because it matches the standard expected at SmartBear to use GitHub Action workflows for CI/CD.

## Changeset

Added a new workflow directory and file to allow us to publish releases tagged to NPM under our SmartBear Account

## Testing

I plan to test using workflow dispatcher and the --dry-run flag once I can see the file in the Actions UI.


.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,36 @@
...
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18' # or whatever version you require
```
sazap10, line 22:
```suggestion
          node-version: '22'
```

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  publish:
+    name: Publish package
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
sazap10, line 12: we shouldn't need this as we are only publishing to npm rather than github packages

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish to npm
+
+on:
+  workflow_dispatch:
```
sazap10, line 4: maybe we can go with:
```suggestion
on:
  release:
    types: [published]
```
so it publishes to npm when a release is published

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          node-version: '18' # or whatever version you require
+          registry-url: 'https://registry.npmjs.org/'
+          scope: '@smartbear'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build package
+        run: npm run build # skip this step if not needed
```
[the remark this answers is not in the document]
sazap10: also, `npm ci` will run a build

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,30 @@
...
+          registry-url: 'https://registry.npmjs.org/'
+          scope: '@smartbear'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Publish to npm
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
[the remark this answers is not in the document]
sazap10: this would be good to add

PR #17, opened 2025-06-20
Read version from package.json and create static server name

sazap10:
## Goal

Allows reading the version from the `package.json`. Allowing it to be used within the server for the MCP server version, User agents, etc

## Design

Requires changing module to target `node18` where reading json via import is supported

## Changeset

Added `common/info.ts` which declares `MCP_SERVER_NAME` and `MCP_SERVER_VERSION` and imported to `index.ts` and into Insight Hub client (to use for user agent).

Also removed `console.info` calls as these are not valid messages to output when running `stdio` server

## Testing

No change to functionality


common/info.ts
```diff
@@ -0,0 +1,3 @@
+import packageJson from '../package.json' with { type: "json" };
+export const MCP_SERVER_NAME = "SmartBear MCP Server";
```
tomlongridge, line 2: What about using `description` from the `package.json` for this, or using a [custom](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#config) `config` entry?

sazap10: yeh, updated to add config to package.json

tomlongridge (approved):

PR #26, opened 2025-07-04
[Insight Hub] Add endpoint config and status code response handling

tomlongridge:
Added logic to handle the new Hub API key.

It also now throws an error if the status code is not successful, rather than going ahead to try and parse the response, so now it gives a more understandable error.

### DIFF
--- a/insight-hub/README.md
+++ b/insight-hub/README.md
@@ -26,9 +26,9 @@ To connect an MCP server, you will need to create a personal auth token from the
    - Get the latest event for an error.
 4. `get_insight_hub_event_details`
    - Get details of a specific event.
-6. `get_project_event_filters`
+5. `get_project_event_filters`
    - List the filters available for a project.
-5. `list_insight_hub_project_errors`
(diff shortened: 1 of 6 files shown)


tomlongridge:
I'll merge to `next` after #27

insight-hub/README.md
```diff
@@ -2,34 +2,37 @@
...
    - List all projects in an organization.
+   - Multi-project mode only.
 2. `get_insight_hub_error`
    - Get error details from a project.
 3. `get_insight_hub_error_latest_event`
    - Get the latest event for an error.
 4. `get_insight_hub_event_details`
-   - Get details of a specific event on Insight Hub.
+   - Get details of a specific event.
+6. `get_project_event_filters`
```
bullmo, line 29: The numbering is out of sync

sazap10: numbering doesnt really matter too much in markdown files but yeh we should fix that

sazap10 (approved):

PR #42, opened 2025-07-18
fix: remove the search field from filtering

tomlongridge:
## Goal

Improve the error filtering experience.

## Changeset

* Removes `search` from all event fields as this matches multiple fields. We'd rather the LLM uses specific fields – e.g. `error.message`.
* Added a prompt hint that filter results may be empty and that's not a failure case – avoids the LLM trying different search approaches.

## Testing

Manual.

### DIFF
--- a/insight-hub/client.ts
+++ b/insight-hub/client.ts
@@ -22,9 +22,9 @@ const cacheKeys = {
 }
 
 // Exclude certain event fields from the project event filters to improve agent usage
-const EXCLUDED_EVENT_FIELDS = [
+const EXCLUDED_EVENT_FIELDS = new Set([
   "search" // This is searches multiple fields and is more a convenience for humans, we're removing to avoid over-matching
-];
+]);
 
 // Type definitions for tool arguments
 export interface ProjectArgs {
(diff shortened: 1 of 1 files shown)


insight-hub/client.ts
```diff
@@ -21,6 +21,11 @@ const cacheKeys = {
   PROJECT_EVENT_FILTERS: "insight_hub_project_event_filters",
 }
 
+// Exclude certain event fields from the project event filters to improve agent usage
+const EXCLUDED_EVENT_FIELDS = [
+  "search" // This is searches multiple fields and is more a convenience for humans, we're removing to avoid over-matching
+];
```
lemnik, line 27: Given how this is used, is it worth using a `Set` instead of an array?
```suggestion
const EXCLUDED_EVENT_FIELDS = new Set([
  "search" // This is searches multiple fields and is more a convenience for humans, we're removing to avoid over-matching
]);
```

If so: `EXCLUDED_EVENT_FIELDS.includes` changes to `EXCLUDED_EVENT_FIELDS.has`

sazap10 (approved):

Issue #59, opened 2025-08-04
Scripts accessing packages that are not listed as dependencies


domarmstrong:
### Describe the bug
When using yarn you can only access packages that are declared as your own dependencies otherwise they are considered 'unsound'.

```
$ yarn dlx @smartbear/mcp@latest

....

Error: @smartbear/mcp tried to access zod, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: zod (via "zod/package.json")
Required by: @smartbear/mcp@npm:0.2.2 (via /Users/dom.armstrong/.yarn/berry/cache/@smartbear-mcp-npm-0.2.2-deaf14e619-10.zip/node_modules/@smartbear/mcp/dist/insight-hub/client.js)
```

PR #60, opened 2025-08-04
Add zod as a dependency

domarmstrong:
## Goal

See [Scripts accessing packages that are not listed as dependencies](https://github.com/SmartBear/smartbear-mcp/issues/59)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
   "dependencies": {
     "@bugsnag/js": "^8.2.0",
     "@modelcontextprotocol/sdk": "^1.15.0",
-    "node-cache": "^5.1.2"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,7 +42,8 @@
   "dependencies": {
     "@bugsnag/js": "^8.2.0",
     "@modelcontextprotocol/sdk": "^1.15.0",
-    "node-cache": "^5.1.2"
+    "node-cache": "^5.1.2",
+    "zod": "^3"
```
domarmstrong, line 46: I left the version loose so we use the same version as the current modelcontextprotocol/sdk package

PR #70, opened 2025-08-08
abstract server and registerTools/Resources

tomlongridge:
## Goal

Provide a better framework for using the McpServer with different tools across different clients (products):

* Consolidates the tool description template
* Surrounds each tool with Bugsnag error monitoring
* Allows for a product name to be used for each client.

## Design

The `SmartBearMcpServer` extends `McpServer` to provide a way to intercept parameters – the server calls each added client to register its tools and resources, providing a function for each tool/resource.


sazap10 (approved):
SFTM

PR #71, opened 2025-08-10
Pact 4423 mcp tool generate pacts

prateek0411999:
This PR introduces the PactFlow Client, enabling Pact test generation via `Pactflow AI` and laying the groundwork for future Pact Broker support.

**Integration of PactFlow Client:**

- `pactflow/client.ts` – Added the PactflowClient class, which provides methods for interacting with PactFlow's APIs. This includes authentication handling, dynamic tool registration with the McpServer, and request configuration.
- `pactflow/client/ai.ts` – Defines interfaces for PactFlow AI routes.
- `pactflow/client/base.ts` – Defines interfaces for shared PactFlow/Pact Broker routes.
- `pactflow/client/tools.ts` – Defines tools for registration, with clientType support for filtering.

**Note:**  Since We’re adding support for both Pactflow and Pact Broker, tools are registered dynamically based on the clientType.


mefellows:
I think we'll need to rebase this change onto https://github.com/SmartBear/smartbear-mcp/pull/70 Prateek.

kevinrvaz (approved):

JP-Ellis (approved):
Approving, though it would be good if the Zod calls could re-use the interfaces/objects defined in TypeScript (or have the TypeScript types defined by Zod?).

Happy to use a follow up PR if need be.

PR #85, opened 2025-08-11
fix manifest for doc publishing

frankkilcommins:
## Goal

To ensure that all docs changes get autopublished.

## Changeset

Adjusted the manifest order to ensure TOC entries are correct. Adjusted the description to match the developer portal product description


frankkilcommins:
@sazap10 you might want to take a look at the test scripts/workflow. I wouldn't expect them to be failing based on the type of change in this PR

sazap10:
> @sazap10 you might want to take a look at the test scripts/workflow. I wouldn't expect them to be failing based on the type of change in this PR

I believe its because its a forked repo which doesnt have access to push to this repo, it works fine in a separate PR. Ill see if we can update it to work for both forked and non forked PRs

frankkilcommins:
@sazap10 thanks for approving. Will let you merge in at your leisure

PR #87, opened 2025-08-19
feat: support for fetching provider-states for pactflow and pact broker

prateek0411999:
This PR adds support for fetching **provider states** via the Pactflow client.

Since the provider-states API is available in both Pactflow and Pact Broker, support for configuring pact broker has been included as well.

Additionally, fetching provider states is now registered as a resource in MCP, enabling clients to access this functionality through the resource interface.

Refactor:
- Registered Provider-States as a tool ( discussion [here](https://github.com/SmartBear/smartbear-mcp/pull/87#discussion_r2287059652) )

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -49,20 +49,21 @@ async function main() {
     client_defined = true;
   }
 
-  if(pactBrokerToken && pactBrokerUrl) {
-    server.addClient(new PactflowClient(pactBrokerToken, pactBrokerUrl, "pactflow"));
-    client_defined = true;
-  }
-
-  if(pactBrokerUrl && pactBrokerUsername && pactBrokerPassword){
-    const pactBrokerClient = new PactflowClient({ username: pactBrokerUsername, password: pactBrokerPassword }, pactBrokerUrl, "pact_broker");
-    server.addClient(pactBrokerClient);
-    client_defined = true; 
+  if (pactBrokerUrl) {
+    if (pactBrokerToken) {
+      server.addClient(new PactflowClient(pactBrokerToken, pactBrokerUrl, "pactflow"));
+      client_defined = true;
+    } else if (pactBrokerUsername && pactBrokerPassword) {
+      server.addClient(new PactflowClient({ username: pactBrokerUsername, password: pactBrokerPassword }, pactBrokerUrl, "pact_broker"));
+      client_defined = true;
+    } else {
+      console.error("If the Pact Broker base URL is specified, you must specify either (a) a PactFlow token, or (b) a Pact Broker username and password pair.")
+    }
   }
 
   if (!client_defined) {
     console.error(
-      "Please set one of REFLECT_API_TOKEN, INSIGHT_HUB_AUTH_TOKEN, API_HUB_API_KEY or PACT_BROKER_TOKEN / PACT_BROKER_BASE_URL environment variables",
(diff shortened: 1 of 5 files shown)


pactflow/client.ts
```diff
@@ -143,4 +162,20 @@ export class PactflowClient implements Client {
         );
       }
     }
+
+    // Register resources for supported clients.
+    // To add a new resource:
+    //   1. Define the list of allowed clients in the `if` condition.
+    //   2. Call `register` with the resource name, path, and handler.
+    //   3. Only clients in the allowed list will have access to that resource.
+    registerResources(register: RegisterResourceFunction): void {
```
mefellows, line 171: I'd be keen for your thoughts on why we chose to use a `resource` instead of a `tool`. Some of the use cases I imagine this being used in:

1. When creating consumer tests to re-use existing states, where possible, from the existing provider
2. When creating provider side tests, using the known states to create state handlers

Ideally, 1 and 2 could be used in prompts or wider agentic use cases, where the agent knows to reach out to this resource/tool to help. For example, if we were to create a new `prompt` to help an agent setup Pact on a project, we'd encourage the use of this resource/tool to help establish the setup.

I'm not sure if a tool or resource is best used here.

pactflow/client.ts
```diff
@@ -143,4 +162,20 @@ export class PactflowClient implements Client {
...
+      if(["pactflow", "pact_broker"].includes(this.clientType)) {
+        register("get-provider-states", "{provider}", async (uri, variables, _extra) => {
+          const states = await this.getProviderStates(variables.provider as string);
+          return {
+            contents: [{ uri: uri.href, text: JSON.stringify(states) }]
```
mefellows, line 176: Should we create a type for this response, to help the agent know what to expect here?

index.ts
```diff
@@ -54,11 +54,11 @@ async function main() {
     client_defined = true;
   }
 
-  // Once PactBroker tools are implemented, we can uncomment this
-  // if(pactBrokerUrl && pactBrokerUsername && pactBrokerPassword){
-  //   const pactBrokerClient = new PactflowClient({ username: pactBrokerUsername, password: pactBrokerPassword }, pactBrokerUrl, "pactbroker");
-  //   server.addClient(pactBrokerClient);
-  // }
+  if(pactBrokerUrl && pactBrokerUsername && pactBrokerPassword){
```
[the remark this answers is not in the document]
kevinrvaz: Should we throw the error instead of adding it to console?, considering its a local mpc server the user will not be able to see this error unless the open app logs

pactflow/client.ts
prateek0411999: Ideally, this should always be implemented as a **resource**. However, when testing things out, GitHub Copilot (in agent mode on VS Code) doesn’t yet support calling resources directly through prompts—it only supports MCP tools. To call a resource, we currently have to go through the command palette.

Agents like Claude can already handle resources, but since VS Code + Copilot are the most widely used, it makes sense to stick with a tool for now. Once Copilot’s agent mode adds support for resources, we can switch back to using a resource.

thoughts @JP-Ellis @mefellows ?

index.ts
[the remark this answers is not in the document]
JP-Ellis: Error would be acceptable if they've defined the URL and nothing else. Happy to defer with the convention used elsewhere.

JP-Ellis (approved):

PR #88, opened 2025-08-22
Add pagination, sorting and total counts to list errors tool

gingerbenw:
## Goal

Adds support for sorting and pagination to the `List Project Errors` tool

## Design

The tool output now includes a `next` URL extracted from the `Link` header that clients can pass back to the tool as a parameter to fetch the next page of results, as well as a separate `total` field (from the `X-Total-Count` header) to distinguish between total results and returned results.

Also added sort and direction parameters to support more flexible searching.

## Testing

Tested manually in vscode


insight-hub/client.ts
```diff
@@ -497,6 +498,28 @@ export class InsightHubClient implements Client {
               "Relative time periods: h (hours), d (days)"
             ]
           },
+          {
```
sazap10, line 501: does it make sense to also provide `per_page` as a parameter rather than getting 50 per page?

yousif-bugsnag: yeah, I switched to a fixed page size because the LLMs seemed to be confused by it, but I think I've resolved that with examples and hints - update incoming...

PR #89, opened 2025-08-22
feat: Added Pactflow AI review tool

kevinrvaz:
Ref: PACT-4494

## Goal

This PR introduces support for the Pactflow AI review functionality as a MCP tool

## Testing
This tool was tested using github copilot as well as MCP inspector

### DIFF
--- a/common/server.ts
+++ b/common/server.ts
@@ -78,7 +78,7 @@ export class SmartBearMcpServer extends McpServer {
 
     private getInputSchema(params: ToolParams): any {
         const args: ZodRawShape = {};
-        for (const param of params.parameters) {
+        for (const param of params.parameters ?? []) {
             args[param.name] = param.type;
             if (param.description) {
                 args[param.name] = args[param.name].describe(param.description);
@@ -87,6 +87,21 @@ export class SmartBearMcpServer extends McpServer {
                 args[param.name] = args[param.name].optional();
             }
         }
+
+        if (params.zodSchema && params.zodSchema instanceof ZodObject) {
+            for (const key of Object.keys(params.zodSchema.shape)) {
+                const field = params.zodSchema.shape[key];
+                args[key] = field;
+                if (field.description) {
+                    args[key] = args[key].describe(field.description);
+                }
+
+                if (field.isOptional()) {
+                    args[key] = args[key].optional();
+                }
+            }
+        }
+
         return args;
     }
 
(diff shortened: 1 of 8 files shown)


common/server.ts
```diff
@@ -96,21 +111,30 @@ export class SmartBearMcpServer extends McpServer {
...
                 `${p.examples ? ` (e.g. ${p.examples.join(', ')})` : ''}` +
                 `${p.constraints ? `\n  - ${p.constraints.join('\n  - ')}` : ''}`
             ).join('\n')}`;
         }
+        
+        if (zodSchema && zodSchema instanceof ZodObject) {
+            description += "\n\n**Parameters:**\n";
+            for (const key of Object.keys(zodSchema.shape)) {
+                const field = zodSchema.shape[key];
+                description += `- ${key} (${this.getReadableTypeName(field)})${field.isOptional() ? '' : ' *required*'}${field.description ? `: ${field.description}` : ''}\n`;
```
prateek0411999, line 135: when using the zodSchema, would we be missing out on the examples and constraints? (In case we’re considering adding those in somehow.)

pactflow/client/ai.ts
```diff
@@ -101,3 +68,107 @@ export interface RefineRecommendation {
...
+    ),
+  body: z
+    .string()
+    .describe("Complete file contents - client code, models, test files, etc."),
+});
+
+export const OpenAPISchema = z
+  .object({
+    openapi: z.string().describe("For OpenAPI version (e.g., '3.0.0')"),
+    swagger: z
```
prateek0411999, line 92: I think we should make both OpenAPI and Swagger optional, and also include the same validation logic we had earlier.

prateek0411999 (commented):
Looks good! Could we also update the Generate Pact tools parameters as part of this change?

common/server.ts
kevinrvaz: while zod by default won't support them, i can add a custom schema for examples and constraints

prateek0411999:
Looks good! Could we also update the Generate Pact tools parameters as part of this change?

> > Looks good! Could we also update the Generate Pact tools parameters as part of this change?
> 
> while initially i was doing that, i then decided i will raise a followup pr for that to keep this pr just for review tool, but i can re-add those, let me know.

We can raise a follow-up PR for that, Thanks.

prateek0411999 (approved):

PR #104, opened 2025-09-01
feat: add remote OAD/OAS support for review and generate tool in PactFlow

kevinrvaz:
Ref: PACT-4255

## Goal

This PR introduces support for remote OAD reading in both the generate and review pact tool (Note that the auth token in screenshots is not a valid one)

## Testing

Changes were tested via MCP inspector and github copilot

### Review Pacts with swagger url with auth token:-
Screenshot 2025-09-01 at 12 36 25 PM

### Generate Pacts with swagger url with auth token:-
Screenshot 2025-09-01 at 12 37 04 PM

### Generate pacts with public github url:-
Screenshot 2025-09-01 at 2 34 36 PM


tests/unit/pactflow/client.test.ts
```diff
@@ -78,10 +96,46 @@ describe("PactflowClient.registerTools", () => {
...
+    });
+    expect(fetch).toHaveBeenCalledWith("https://api.example.com/openapi.json", {
+      headers: {
+        Authorization: "Bearer my-token",
+      },
+      method: "GET",
+    });
+  });
+
+  it("Gets throws an error if url is missing", async () => {
```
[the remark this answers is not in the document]
kevinrvaz: updated

mefellows:
Rather than providing the token in the chat, would an [elicitation](https://modelcontextprotocol.io/specification/draft/client/elicitation) be a better option?

mefellows:
> Rather than providing the token in the chat, would an [elicitation](https://modelcontextprotocol.io/specification/draft/client/elicitation) be a better option?

Nevermind, that's explicitly not meant to happen:
Screenshot 2025-09-02 at 11 13 19 am

pactflow/client/ai.ts
```diff
@@ -131,9 +151,34 @@ export const EndpointMatcherSchema = z
...
+      .url("Must be a valid openapi url")
+      .describe("URL of the remote OpenAPI document.")
+      .optional(),
+  })
+  .describe("Use this schema to fetch openapi documents present over a url.");
+
 export const OpenAPIWithMatcherSchema = z
   .object({
-    document: OpenAPISchema,
+    document: OpenAPISchema.describe(
```
JP-Ellis, line 179: Just a minor nit: Don't be afraid to raise separate PRs for logically distinct changes. The additional descriptions here are good to include, but are not actually relevant to the remote OAD support.

pactflow/client/ai.ts
```diff
@@ -85,19 +85,39 @@ export const OpenAPISchema = z
...
+          .optional(),
+        license: z
+          .object({
+            name: z.string().describe("License name"),
+            url: z.string().url().describe("License URL"),
+          })
+          .optional(),
+      })
+      .describe("Information about the API")
+      .optional(),
```
JP-Ellis, line 108: It isn't clear to me why we need to check for this information. I don't think we use this anywhere, in which case, we should just transparently pass it through.

pactflow/client.ts
```diff
@@ -42,16 +54,31 @@ export class PactflowClient implements Client {
...
+     * @throws Error if the HTTP request fails or the operation times out.
+     */
+    async generate(toolInput: GenerationToolInput): Promise<GenerationResponse> {
       // Submit the generation request
+      if (toolInput.remoteOpenAPIDocument) {
+        const resolvedSpec = await this.resolveOpenAPISpec(toolInput.remoteOpenAPIDocument);
+        if (toolInput.body.openapi) {
+          toolInput.body.openapi.document = resolvedSpec;
+        } else {
+          throw new Error("When providing a remote OpenAPI document, the 'openapi' field must atleast have a matcher if remote openapi credentials are used.");
```
JP-Ellis, line 70: This error message doesn't fully make sense to me. (Also, minor typo)


```suggestion
          throw new Error("When providing a remote OpenAPI document, the 'openapi' field must at least have a matcher if remote OpenAPI credentials are used.");
```

pactflow/client/ai.ts
```diff
@@ -200,6 +245,24 @@ export const GenerationInputSchema = z.object({
...
 });
 
+export const GenerationToolInputSchema = z.object({
+  body: GenerationInputSchema.describe(
+    "The body of the request to be used for Pact test generation."
+  ),
+  remoteOpenAPIDocument: RemoteOpenAPIDocumentSchema.optional().describe(
+    "The remote OpenAPI document to use for the review in case openapi document is not provided. If provided do not include the document field under openapi."
+  ),
+});
```
JP-Ellis, line 255: This feels somewhat incorrect as a type. From what I understand, the input should _either_ be a local file _or_ a remote file. This would imply that you should use an enum or some `A | B` type. What you have you imply that you can have both a `body` and a `remoteOpenAPIDocument`.

pactflow/client.ts
```diff
@@ -61,16 +88,90 @@ export class PactflowClient implements Client {
...
+        const resolvedSpec = await Swagger.resolve({ spec: unresolvedSpec });
+
+        if (!resolvedSpec.errors?.length) {
+          return resolvedSpec.spec;
+        }
+
+        throw new Error(`Failed to resolve OpenAPI document: ${resolvedSpec.errors?.join(", ")}`);
+      }
+
+      throw new Error(`Invalid RemoteOpenAPIDocument: ${JSON.stringify(openAPISchema.error?.issues)}`);
```
JP-Ellis, line 111: This is a minor nit, but I think it is nicer to have early exits than repeated nested conditions.

E.g.:

```
if (!openAPISchema.success) {
   throw new ...
}

// now we know OpenAPI schema was retrieved.

if (resolvedSpec.errors) {
  throw new ...
}

// Now we know there are no errors
```

I don't code in TypeScript all that often, so what I'm suggesting is contingent on TypeScript's inferences working as I expect.

pactflow/client.ts
```diff
@@ -61,16 +88,90 @@ export class PactflowClient implements Client {
...
+          case "application/json":
+            return await remoteSpec.json();
+          case "text/plain":
+          case "text/plain;charset=utf-8":
+          case "text/plain; charset=utf-8":
+          case "application/yaml":
+            return JSON.parse(JSON.stringify(yaml.load(await remoteSpec.text()), null, 2));
+          default:
+            throw new Error(`Unsupported Content-Type: ${remoteSpec.headers.get("Content-Type")}`);
+        }
```
JP-Ellis, line 144: I would not match the content type this way as you'll be bound to miss something. For example:

- Other charsets: `application/json; charset=ascii`, `application/json;charset=utf-16`, etc.
- Custom content types: `application/vnd.custom.openapi+json`, `application/vnd.custom.openapi+yaml`, or maybe even `application/openapi+json`

Instead:

1. I would trust that `charset` to be used to parse the input to text correctly by the underlying library
2. I would try to deserialise as JSON first, and automatically fallback to YAML irrespective of the content type. 
3. If both deserialisations fail, then throw an error.

pactflow/client.ts
```diff
@@ -61,16 +88,90 @@ export class PactflowClient implements Client {
...
+      }
+
+      throw new Error("'url' must be provided.");
```
JP-Ellis, line 147: As above, I think early exits are nicer, but that's really just a stylistic preference.

JP-Ellis (changes requested):
Looking good, though there are a few things I think need to be changed.

I think the main thing is how the input is given, and I'm wondering how much we should rely on the LLM to figure out, vs us handling automatically.

I'm wondering whether the `OpenApiWithMatcher` should be adjusted to be:

```
OpenApiWithMatcher:
   matcher: Matcher
   openapi: URI | OpenApi

URI:
   location: str
   headers: list[tuple[str, str]] // For custom headers
   auth: tuple[str, str] // For the authorisation header
```

and then if we have a URI, then when it is processed, the URI gets resolved into an OpenAPI doc

pactflow/client/ai.ts
kevinrvaz: This text has been kept just to nudge the LLM to ensure that the document field doesn't get added during creation of tool params as discussed offline.

pactflow/client/ai.ts
kevinrvaz: This has been added as this structure can come as part of a oad which may contain it, and can occasionally be malformed, in which case a zod error will be thrown and the llm will then prompt the user to fix the issues in the oad before proceeding

JP-Ellis: But since we're feeding this to an LLM, we don't actually need a fully spec-compliant OAD (in fact, server side, I don't think we check the structure of this at all). If this is somehow malformed, the LLM will likely recover.

pactflow/client.ts
kevinrvaz: as discussed offline i will change this with a transform/refine operation over zod

pactflow/client.ts
```diff
@@ -61,16 +89,90 @@ export class PactflowClient implements Client {
...
+      
+      const unresolvedSpec = await this.getRemoteSpecContents(openAPISchema.data);
+      const resolvedSpec = await Swagger.resolve({ spec: unresolvedSpec });
+
+      if (resolvedSpec.errors) {
+          throw new Error(`Failed to resolve OpenAPI document: ${resolvedSpec.errors?.join(", ")}`);
+      }
+
+      return resolvedSpec.spec;
+    }
```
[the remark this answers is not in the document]
kevinrvaz: updated

pactflow/client.ts
kevinrvaz: updated.

pactflow/client/ai.ts
```diff
@@ -131,14 +131,42 @@ export const EndpointMatcherSchema = z
...
+    ).optional(),
     matcher: EndpointMatcherSchema,
+    remoteDocument: RemoteOpenAPIDocumentSchema.optional().describe(
+      "The remote OpenAPI document to use for the review/generation in case openapi document is not provided. If provided do not include the document field under openapi."
+    ),
   })
   .describe(
     "If provided, the OpenAPI document which describes the API being tested and is accompanied by a matcher which will be used to identify the interactions in the OpenAPI document which are relevant to the Pact refinement process."
-  );
+  ).transform(addOpenAPISpecToSchema);
```
JP-Ellis, line 169: Nice! I think that's a much better solution as it ensures we have a valid structure all the time 🚀 I can see the benefit in where it is used, the code is a lot cleaner.

pactflow/client/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+    try {
+      return JSON.parse(JSON.stringify(yaml.load(specRawBody), null, 2));
+    } catch (e) {
+      throw new Error(
+        `Unsupported Content-Type: ${remoteSpec.headers.get(
+          "Content-Type"
+        )} for remote OpenAPI document. ${e}`
+      );
+    }
+  }
```
JP-Ellis, line 85: Last change I would like to see here: can you catch both the JSON and YAML errors? If both fail and you throw an error, you can list both of them and the end-user can determine which one is valid based on what they are expecting.

pactflow/client/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+    method: "GET",
+  });
+
+  const specRawBody = await remoteSpec.text();
+
+  try {
+    return JSON.parse(specRawBody);
+  } catch {
+    try {
+      return JSON.parse(JSON.stringify(yaml.load(specRawBody), null, 2));
```
JP-Ellis, line 77: It feels like you're just doing $f^{-1}(f(x)) = x$ here (i.e., `stringify` and `parse` are inverses of each other, and therefore their composition results in the identity function $I : x \mapsto x$).

Why couldn't you have:


```suggestion
      return yaml.load(specRawBody);
```

JP-Ellis (changes requested):
Two last changes and I think we'll be good to merge 🚀

pactflow/client/utils.ts
kevinrvaz: The Swagger-js lib can only parse a JS object, which is why even the yaml needs to be converted into a json and returned back.

JP-Ellis (approved):
🚀

PR #106, opened 2025-09-01
feat: add support for can-i-deploy in pactflow and pact_broker client

prateek0411999:
## Goal

This PR introduces support for `can-i-deploy` in PactFlow and Pact Broker client, enabling users to verify if a pacticipant version is safe to deploy.

## Testing

- Valid Input:

Screenshot 2025-09-02 at 1 45 09 AM

Screenshot 2025-09-02 at 1 32 55 AM

- Invalid Input ( When an Invalid Pacticipant is provided ) 
Screenshot 2025-09-02 at 1 35 09 AM

### DIFF
--- a/pactflow/client.ts
+++ b/pactflow/client.ts
@@ -200,8 +200,7 @@ export class PactflowClient implements Client {
 
     registerTools(register: RegisterToolsFunction, _getInput: GetInputFunction): void {
       for (const tool of TOOLS.filter(t => t.clients.includes(this.clientType))) {
-        const {handler, clients, formatResponse, ...toolparams} = tool;
-        void clients;
+        const {handler, clients: _, formatResponse, ...toolparams} = tool; // eslint-disable-line @typescript-eslint/no-unused-vars
         register(toolparams, async (args, _extra) => {
             const handler_fn = (this as any)[handler];
             if (typeof handler_fn !== "function") {
--- a/pactflow/client.ts
+++ b/pactflow/client.ts
@@ -2,7 +2,7 @@ import { GenerationInput, GenerationResponse, RefineInput, RefineResponse, Statu
 import { ClientType, TOOLS } from "./client/tools.js";
 import { MCP_SERVER_NAME, MCP_SERVER_VERSION } from "../common/info.js";
 import { Client, GetInputFunction, RegisterToolsFunction } from "../common/types.js";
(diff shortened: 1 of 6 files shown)


mefellows:
We should also please update the docs as part of this (i.e. what goes here: https://developer.smartbear.com/smartbear-mcp/docs/contract-testing-with-pactflow).

pactflow/client.ts
```diff
@@ -156,10 +156,52 @@ export class PactflowClient implements Client {
...
+        return (await response.json()) as CanIDeployResponse;
+      } catch (error) {
+        console.error("[CanIDeploy] Unexpected error:", error);
+        throw error;
+      }
+    }
+
     registerTools(register: RegisterToolsFunction, _getInput: GetInputFunction): void {
       for (const tool of TOOLS.filter(t => t.clients.includes(this.clientType))) {
         const {handler, clients, formatResponse, ...toolparams} = tool;
```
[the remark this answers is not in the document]
prateek0411999: We can’t omit properties during destructuring directly, but we can rename them to `_` like this 
`const {handler, clients: _, formatResponse, ...toolparams} = tool;` 

However, this will trigger the ESLint warning
'_ 'is assigned a value but never used, so we’d still need to suppress it.

JP-Ellis (approved):
Seems all good to me! Are there any other places to add docs? Or maybe explanations on how to use this?

tests/unit/pactflow/client.test.ts
```diff
@@ -2,86 +2,190 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
...
+
+  describe("registerTools", () => {
+    const mockRegister = vi.fn();
+    const mockGetInput = vi.fn();
+
+    it("registers only tools matching the given clientType", () => {
+      // Arrange — mock TOOLS with multiple client types
+      const fakeTools = [
+        {
+          title: "tool1",
```
kevinrvaz, line 57: should we also add some tests for zodschema? consider we have more zodschema uses over parameters

prateek0411999: We don’t necessarily need both for testing the client types, but I agree that for upcoming tests, incorporating zodSchema would be a good idea.

kevinrvaz (approved):

PR #108, opened 2025-09-02
Ensure test and config files are not published 

yousif-bugsnag:
## Goal

I noticed that we're inadvertently publishing unit test and configuration files because they're being included in the build output. This PR updates the project structure and typescript config to prevent this from happening.

## Design

- Moved the source files and tests into a new `src` directory and made this the root directory for the typescript build
- Updated the tsconfig to exclude test files

## Testing

Manually tested the output using `npm pack`


result.json
sazap10, line 1: accidental checkin?

yousif-bugsnag: woops!

JP-Ellis (approved):
Nice, and helps declutter the root directory 🚀

PR #111, opened 2025-09-02
chore: add pact contract testing for clients

sazap10:
## Goal

Adds contract testing for the various clients

## Design

I used the PactFlow mcp tools to generate and review the contract testing via Copilot

## Changeset

Added contract tests for all four clients
Modified some clients for the pact tests

## Testing

Ran tests locally


tests/contract-testing/api-hub.pact.test.ts
```diff
@@ -0,0 +1,519 @@
...
+            status: 200,
+            headers: {
+              'Content-Type': 'application/json',
+            },
+            body: like(eachLike({
+              id: uuid('123e4567-e89b-12d3-a456-426614174000'),
+              name: string('Test Portal'),
+              subdomain: string('test-portal'),
+              description: string('A test portal for API documentation'),
+              created_at: string('2024-01-01T00:00:00Z'),
```
prateek0411999, line 39: I think we should update the tests to consistently use the appropriate Pact V3 matchers.

For example, use the timestamp matcher for all timestamp fields, and the boolean matcher for boolean fields.

We can refer to Pact V3 matchers [here](https://docs.pact.io/implementation_guides/javascript/docs/matching#v3-matching-rules).

prateek0411999:
Just a note — once this is in place, we’ll also need to implement the provider side. That means verifying the generated Pact contracts against the locally running provider, or alternatively, we could consider running BDCT.

For the Pactflow client, we’re still missing tests for a few endpoints. I think we’ll also need to set up separate tests, since some API calls hit pactflow-ai and others hit pactflow-application (Ruby backend).

We should also consider setting up CI to publish the generated Pact tests to [smartbear.pactflow.io](https://smartbear.pactflow.io/)
 as part of this workflow.

mefellows:
Hey folks,

Sorry for weighing in so late - I thought I had commented already. The PactFlow team will happily pick up the work here to get the initial set of Pact tests setup, published to smartbear.pactflow.io and integrated into the CI pipeline (GitHub). Stay tuned for updates from us in the next week or so.

mefellows:
> Thanks, i was a little lost on the verification side of things. Feel free to start from scratch if thats easier.

No problems! We'll start by getting the consumer tests setup across the board (ideally using MCP to help ;) ) and then we'll get it integrated into CI/CD, and setup a provider verification for the PactFlow App.

Then I reckon we get together as a group to work out how to gradually add the other providers (e.g. Bugsnag, Portal) into the mix.

PR #112, opened 2025-09-03
update changelog and docs for PactFlow

kevinrvaz:
## Goal

Update CHANGELOG.md and contract-testing-with-pactflow.md

### DIFF
--- a/docs/products/SmartBear MCP Server/contract-testing-with-pactflow.md
+++ b/docs/products/SmartBear MCP Server/contract-testing-with-pactflow.md
@@ -1,50 +1,64 @@
 ![api-hub.png](./images/embedded/api-hub.png)
 
 The PactFlow Contract testing client provides comprehensive tools which makes testing and deploying microservices at scale, simple and worry free for thousands of developers and testers around the world. Tools for PactFlow requires a `PACT_BROKER_BASE_URL` along with either a `PACT_BROKER_TOKEN` or (`PACT_BROKER_USERNAME` and `PACT_BROKER_PASSWORD`).
+
 Read more on PactFlow [Docs](https://docs.pactflow.io/).
 
 ## Available Tools
 
-### Generate Pact Tests(`generate`)
+### Generate Pact Tests
 
-- Purpose: Generate Pact tests using an OpenAPI spec in 6 different languages.
+- Purpose: Generate Pact tests from a number of inputs.
 - Returns: Generated Pact test
 - Parameters:
-
-| Parameter |               | Description | Type | Required |
-| --- | --- | --- | --- | --- |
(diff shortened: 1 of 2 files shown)


docs/products/SmartBear MCP Server/contract-testing-with-pactflow.md
```diff
@@ -1,25 +1,54 @@
 ![api-hub.png](./images/embedded/api-hub.png)
 
 The PactFlow Contract testing client provides comprehensive tools which makes testing and deploying microservices at scale, simple and worry free for thousands of developers and testers around the world. Tools for PactFlow requires a `PACT_BROKER_BASE_URL` along with either a `PACT_BROKER_TOKEN` or (`PACT_BROKER_USERNAME` and `PACT_BROKER_PASSWORD`).
-
+Read more on PactFlow [Docs](https://docs.pactflow.io/).
```
JP-Ellis, line 4: If you want this on a new paragraph, you'll need an empty line above:


```suggestion

Read more on PactFlow [Docs](https://docs.pactflow.io/).
```

JP-Ellis (changes requested):
There are some readability tweaks I think are necessary. Since the docs are for human consumption, they need not be overly descriptive. Especially in the context of an MCP server, the LLM can fill in the gaps between the high-level intent from the developer, and the low-level specification/schema and how to invoke which tool.

docs/products/SmartBear MCP Server/contract-testing-with-pactflow.md
```diff
@@ -2,24 +2,74 @@
...
 ## Available Tools
 
 ### Generate Pact Tests
 
-- Generate Pact tests using an OpenAPI spec in 6 different languages.
+- Purpose: Generate Pact tests from a number of inputs.
+- Returns: Generated Pact test
+- Parameters:
+  - **`language`** (optional)
+  The target language in which the Pact tests should be generated. If absent, the language will be inferred from other inputs, such as any provided code files. This field is optional if and only if it can be inferred from other inputs
```
JP-Ellis, line 15: Last nit, but can you give this more breathing room? When rendered, it is very cramped. You can have paragraphs within dot points by having blank lines, and ensuring the next paragraph starts at the same indentation as the start of the text after the dot point marker:


```suggestion
  - **`language`** (optional)
    
    The target language in which the Pact tests should be generated. If absent, the language will be inferred from other inputs, such as any provided code files. This field is optional if and only if it can be inferred from other inputs
```

JP-Ellis (approved):
One last change which I think would help, and it's just stylistic 😅 

Approving now as I don't think it'll require another back-and-forth 🚀

PR #130, opened 2025-09-16
Pact 4685 request matcher if not provided

kevinrvaz:


kevinrvaz:
> If only one matcher is detected, does it still elicit to the user or just default to the obvious one?

can you give an example, i didn't understand the question.

src/common/types.ts
```diff
@@ -34,6 +51,12 @@ export type RegisterResourceFunction = (
     cb: ReadResourceTemplateCallback
 ) => RegisteredResourceTemplate;
 
+export type RegisterPromptFunction = <Args extends ZodRawShape>(name: string, config: {
+  title?: string;
+  description?: string;
+  argsSchema?: Args;
+}, cb: PromptCallback<Args>) => RegisteredPrompt;
```
JP-Ellis, line 58: Just double checking, are all fields genuinely optional?

src/pactflow/client/prompts.ts
```diff
@@ -0,0 +1,166 @@
...
+      ),
+    operationId: z
+      .string()
+      .describe(
+        "OpenAPI operation ID to match (e.g., 'getUserById', 'get*'). Supports glob patterns"
+      ),
+  })
+  .describe(
+    "REQUIRED: Matcher to specify which endpoints from the OpenAPI document to generate tests for. At least one matcher field must be provided"
+  );
```
JP-Ellis, line 70: Wouldn't it be possible to insert the schema here instead of hard-coding code that exists elsewhere in the repository?

src/pactflow/client/prompts.ts
```diff
@@ -0,0 +1,166 @@
...
+            "format": "int64"
+          },
+          "name": {
+            "type": "string"
+          }
+        }
+      }
+    }
+  }
+}
```
JP-Ellis, line 123: As above, I think it would be better to define this in actual TypeScript first, and then insert the JSON-ified string here.

src/pactflow/client/prompts.ts
```diff
@@ -0,0 +1,166 @@
...
+    "statusCodes": [200],
+    "operationId": "getAllUsers"
+  },
+  {
+    "path": "/users/**",
+    "methods": ["GET"],
+    "statusCodes": ["2XX", 404],
+    "operationId": "get*"
+  }
+]
```
JP-Ellis, line 154: And as before for JSON-ified strings

src/pactflow/client/utils.ts
```diff
@@ -99,3 +106,119 @@ export async function addOpenAPISpecToSchema(inputSchema: any) {
...
+        content: {
+          type: "text",
+          text: OADMatcherPrompt.replace("{0}", JSON.stringify(openAPI)),
+        },
+      },
+    ],
+    maxTokens: 1000,
+  });
+
+  const regex = /```json([\s\S]*?)```/;
```
JP-Ellis, line 135: Given that `\S` is the complement of `\s`, then `[\s\S]` will match any character. So it would be more straightforward to have `.*` instead.

```suggestion
  const regex = /```json(.*?)```/;
```

There's also the slight chance of having `json5`, `jsonc`: ```json[c5]?(.*?)```

and there's also the slight chance of different cases. So I would do a case-insensitive search/match.

src/pactflow/client.ts
```diff
@@ -57,10 +76,17 @@ export class PactflowClient implements Client {
      * Generate new Pact tests based on the provided input.
      *
      * @param toolInput The input data for the generation process.
+     * @param getInput Function to get additional input from the user if needed.
      * @returns The result of the generation process.
      * @throws Error if the HTTP request fails or the operation times out.
      */
-    async generate(toolInput: GenerationInput): Promise<GenerationResponse> {
+    async generate(toolInput: GenerationInput, getInput: GetInputFunction): Promise<GenerationResponse> {
+      if(toolInput.openapi?.document && (!toolInput.openapi?.matcher || typeof toolInput.openapi?.matcher === "object" && Object.keys(toolInput.openapi.matcher).length === 0)) {
```
JP-Ellis, line 84: Is the `typeof ... === "object"` necessary? Isn't the matcher guaranteed to be either `null`, `undefined` or an object?

JP-Ellis (approved):
There's a few minor changes, but approving anyway as I doubt it'll require much back-and forth.

src/common/types.ts
kevinrvaz: i've taken this type declaration from one defined in the typescript-mcp sdk but not exported out.

src/pactflow/client.ts
kevinrvaz: these two condition are chained ->
typeof toolInput.openapi?.matcher === "object" && Object.keys(toolInput.openapi.matcher).length === 0

as in some cases the llm might pass an empty object, so if there is an empty object or null or undefined then this condition triggers

JP-Ellis: Yeah, I get that. It's a lazy evaluation:

1. `!toolInput.openapi?.matcher`: Checks that the matcher exists.
2. If it exists, checks its an `object` type
3. If it is an `object` type, check if it is an empty object

But I guess my question was more: If (1) passes (i.e., the matcher field exists / is not null), doesn't the schema guarantee that the matcher is an object already? And `Zod` validates this at runtime?

And on second though, could the check be `toolInput.openapi?.matcher === {}`?

kevinrvaz: on checking this doesnt work with both == and ===, i suppose the zod schema would guarentee its an object so ill remove the object check.

JP-Ellis: Haha, of course `==` and `===` behave weirdly on JS/TS 😆

mefellows:
> > If only one matcher is detected, does it still elicit to the user or just default to the obvious one?
> 
> can you give an example, i didn't understand the question. by detection are we talking about when only 1 matcher is recommended or when user provides one matcher in their initial generate/review prompt?

Yes, when the set of matchers is return with only 1 option - does it elicit a response from the user, or only in the case where there are multiple choices?

kevinrvaz:
> > > If only one matcher is detected, does it still elicit to the user or just default to the obvious one?
> > 
> > 
> > can you give an example, i didn't understand the question. by detection are we talking about when only 1 matcher is recommended or when user provides one matcher in their initial generate/review prompt?
> 
> Yes, when the set of matchers is return with only 1 option - does it elicit a response from the user, or only in the case where there are multiple choices?

you elicit a response even then, as the user can still cancel to provide his own input.

PR #136, opened 2025-09-19
chore: replace eslint+prettier with biomne

JP-Ellis:
## Summary

Add Biome for formatting and linting.

## Motivation

The key benefit (in my experience) of Biome is that it is incredibly fast at formatting and linting, providing near immediate feedback while coding for any lint violations.

The one downside is the Biome plugins are still limited, but this project does not rely heavily on plugins.

I have split the PR into three commits:

- The updates to the config
- Applying automatic formatting and lint fixes
- Applying some non-automatic fixes

## Context

This is an alternative implementation to: 

- #134

I’m happy to discuss or close this if the team prefers the ESLint + Prettier approach. Just wanted to provide a concrete alternative for consideration!


sazap10 (approved):
The github CI changes seem good to me, i will leave the opinion on switching to biome to @gingerbenw and/or @yousif-bugsnag to comment on

PR #143, opened 2025-09-22
Unified publication process

maciejmesjasz:
## Goal

Unify the publication process into a single workflow. It includes: npm, Docker (ECR), GitHub MCP registry.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -24,17 +24,17 @@ jobs:
           node-version: '22'
           registry-url: 'https://registry.npmjs.org/'
           scope: '@smartbear'
-      
+
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@a03048d87541d1d9fcf2ecf528a4a65ba9bd7838 # v5.0.0
         with:
           role-to-assume: ${{ secrets.ECR_IAM_ROLE_ARN }}
           aws-region: us-east-1
-      
+
       - name: Login to Amazon ECR private
         uses: aws-actions/amazon-ecr-login@062b18b96a7aff071d4dc91bc00c4c1a7945b076 # v2.0.1
         with:
-          registries: "709825985650" 
+          registries: "709825985650"
 
       - name: Extract version from package.json
         id: extract_version
@@ -55,7 +55,7 @@ jobs:
           DEST=${{ secrets.ECR_PRIVATE_REGISTRY_URL }}:${IMAGE_TAG}
           docker build -t $DEST .
           docker push $DEST
-          
+
       - name: Install dependencies
         run: npm ci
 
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,87 @@
...
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Build mcp-publisher
+        run: |
+          git clone https://github.com/modelcontextprotocol/registry
+          cd registry
+          make publisher
+          cp bin/mcp-publisher "$GITHUB_WORKSPACE"/mcp-publisher
```
sazap10, line 78: we should be able to replace this by downloading the binary from the latest github release

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,87 @@
...
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    name: Publish package
+    runs-on: ubuntu-24.04
+
```
JP-Ellis, line 16: We should make sure to use an [environment](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments). These are used in a number of places within GitHub to track what has been released. In addition, there are a few additional benefits:

1. We can setup rules as to what can/cannot be deployed to the environment (e.g., prevent pushing a release that has broken CI)
2. We can setup rules that require certain approvals
3. Last, and certainly not least, we can restrict authentication with 3rd parties to that environment only.

That last point I think is quite important because we are using OIDC to authenticate with 3rd parties (NPM, AWS), but it looks like this can be done from any environment... so a malicious or inadvertent PR could authenticate with these third parties.

JP-Ellis (changes requested):
I think we need to make use of environments to help secure things.

Also, the old `publish-npm.yml` workflow should be remove given the new one supersedes it.

.github/workflows/publish.yaml
tomlongridge: Agreed – the docs-as-code process uses this already to have different Portal keys for `Production` (i.e. main) and `Staging` (i.e. not-main).

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,77 @@
...
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    name: Publish package
+    runs-on: ubuntu-24.04
+    environment: Production
```
JP-Ellis, line 16: Nice!

Can you (or anyone with requisite permissions) double check that the secrets are now in the `Production` environment and no longer available in the default CI environment?

Specifically:

- `ECR_IAM_ROLE_ARN`
- `ECR_PRIVATE_REGISTRY_URL`
- `GH_MCP_REGISTRY_KEY`

And also making sure that the various OIDC connections only work from the `Production` environment? From what I can tell, they are:

- NPM
- AWS
- MCP registry

JP-Ellis (approved):
Nice! From the code, it looks good.

Just a question making sure the secrets have been appropriately been moved over, but I can't validate that.

.github/workflows/publish.yaml
tomlongridge: @maciejmesjasz – I think we need to add the NPM key to the environment settings and remove them from the repo-wide list? Then it should all be locked down to just running from main.

maciejmesjasz: Moved to Production environment.

.github/workflows/publish.yaml
maciejmesjasz: Addressed

tomlongridge (approved):

PR #157, opened 2025-09-30
feat: introduced ToolError to improve error handling/reporting

tomlongridge:
## Goal

Standardise error handling by differentiating between errors in the tools that are passed back to the LLM/user and unexpected errors, that are also reported to BugSnag.

The MCP Node framework already converts thrown errors to a standard response with `isError` set to `true`, but this change allows us to catch this earlier and report accordingly.

## Design

Created `ToolError` to be thrown for any error that should just be sent back to the user/LLM and not reported as unexpected.

## Testing

Unit tests updated accordingly and confirmed Bugsnags get reported for unexpected errors.

### DIFF
--- a/src/bugsnag/client/api/base.ts
+++ b/src/bugsnag/client/api/base.ts
@@ -1,3 +1,4 @@
+import { ToolError } from "../../../common/types.js";
 import type { Configuration } from "./configuration.js";
 
 export interface ApiResponse<T> {
@@ -166,9 +167,6 @@ export class BaseAPI {
     let nextUrl: string | null = url;
     let apiResponse: ApiResponse<T[]>;
     do {
-      if (!this.configuration.basePath) {
-        throw new Error("Base path is not configured for API requests");
-      }
       nextUrl = ensureFullUrl(nextUrl, this.configuration.basePath);
       const response: Response = await fetch(nextUrl, {
         ...options,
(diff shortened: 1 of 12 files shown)


src/pactflow/client/prompt-utils.ts
```diff
@@ -1,5 +1,5 @@
 import type { Server } from "@modelcontextprotocol/sdk/server/index.js";
-import type { GetInputFunction } from "../../common/types.js";
+import { type GetInputFunction, ToolError } from "../../common/types.js";
```
sazap10, line 2: should type be outside the braces?

gingerbenw: you can use it either way - inline type imports were added in TypeScript 4 (3.9?) it allows you to import a type alongside actual code on one line. (we're importing `ToolError` not as a type)

sazap10: ah fair enough, i clearly dont know enough about typescript 😅

src/bugsnag/client/api/base.ts
```diff
@@ -87,14 +88,11 @@ export class BaseAPI {
     let nextUrl: string | null = options.url;
     let apiResponse: ApiResponse<T>;
     do {
-      if (!this.configuration.basePath) {
-        throw new Error("Base path is not configured for API requests");
-      }
       nextUrl = ensureFullUrl(nextUrl, this.configuration.basePath);
```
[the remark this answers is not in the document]
tomlongridge: It's now a non-null field, so the check isn't required.

sazap10 (approved):

PR #162, opened 2025-10-01
chore: add typos to pre-commit checks

JP-Ellis:
The [typos](https://github.com/crate-ci/typos) CLI is a great spell checker for the source code. For example, catching misspellings of function or variable names, or comments.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,8 @@ repos:
     rev: v2.2.4
     hooks:
     - id: biome-check
+
+  - repo: https://github.com/crate-ci/typos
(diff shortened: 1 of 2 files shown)


JP-Ellis:
The problem with standard spell checking tools is that they are design to spell English text; whereas `typos` is specifically design to work with code. For example, it will suggest fixes for things like:

```diff
- addNewSelctor
+ addNewSelector

- const my_new_varible = ...;
+ const my_new_variable = ...;
```

I see this as being complementary to the check we do with cspell. In particular, we use cspell to validate the docs (and we could even augment that to add some grammar checking), whereas typos is used to check the code itself.

PR #168, opened 2025-10-03
chore: improve pactflow ai status description and use-cases

prateek0411999:
## Goal

This PR updates the Description and Use Cases for the PactFlow AI status to improve clarity, ensuring the agent only calls getAIStatus for 401 Unauthorized errors and not for other types of API errors.

### DIFF
--- a/src/pactflow/client/tools.ts
+++ b/src/pactflow/client/tools.ts
@@ -90,19 +90,19 @@ export const TOOLS: PactflowToolParams[] = [
     clients: ["pactflow", "pact_broker"],
   },
   {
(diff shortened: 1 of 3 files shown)


src/pactflow/client/tools.ts
```diff
@@ -92,15 +92,15 @@ export const TOOLS: PactflowToolParams[] = [
   {
     title: "PactFlow AI Status",
```
kevinrvaz, line 93: can we also change the title?

prateek0411999: yeah

PR #171, opened 2025-10-03
[Zephyr] TM4J-13478 Implement `Get Projects` tool and initial foundation

assisblf:
## Goal

Implement one Zephyr tool and lay the foundation for upcoming tools.

## Design

* Put tool definitions and callbacks in dedicated classes for easier readability/maintenance
* Use types and schemas to avoid code duplication while keeping clarity

## Changeset

* Added instructions/documentation
* Created an API client to be used in other Zephyr tools
* Added the tool `zephyr_get_projects`

## Testing

* Added unit tests
* Tested the functionality using GitHub Copilot and invoking the tool


juro994:
Shouldn't we also change the general README [here](https://github.com/SmartBear/smartbear-mcp/blob/7d59cf749133746454734081f38b353752ebfdc0/README.md?plain=1#L69), where the installation instructions are, by adding Zephyr specific environment variables?

src/tests/unit/zephyr/common/api-client.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  it("should build correct URL with params", () => {
+    const url = apiClient.getUrl("/projects", { maxResults: 5, startAt: 1 });
+    expect(url).toBe(`${baseUrl}/projects?maxResults=5&startAt=1`);
+  });
+
+  it("should build correct URL with undefined params", () => {
+    const url = apiClient.getUrl("/projects", { maxResults: undefined });
+    expect(url).toBe("https://api.zephyrscale.smartbear.com/v2/projects");
```
juro994, line 25: Here, we should probably use the interpolation with `${baseUrl}`, like in the test case above. Right?

juro994 (approved):
Nice job!

I've just left 2 very small comments.

I've tested your changes by installing the server locally in VSCode and it looks great 🚀

src/zephyr/tool/get-projects.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+      {
+        name: "startAt",
+        type: StartAtSchema,
+        required: false,
+      },
+    ],
+  };
+
+  handle: ToolCallback<{ [key: string]: any }> = async (args, _extra) => {
```
tomlongridge, line 30: The Pact team have a nice trick [here](https://github.com/SmartBear/smartbear-mcp/blob/3e87ebfa74cb917e8e1dfd35f755e68e1b28cc17/src/pactflow/client/base.ts#L128) to infer the type from the Zod schema – possibly something to use here so you have matching types in the callback.

tomlongridge (commented):
Can you update the README and add a CHANGELOG entry for this?

src/zephyr/tool/get-projects.ts
```diff
@@ -0,0 +1,40 @@
...
+import type { ZephyrTool } from "./zephyr-tool.js";
+
+export class GetProjects implements ZephyrTool {
+  private readonly apiClient: ApiClient;
+
+  constructor(apiClient: ApiClient) {
+    this.apiClient = apiClient;
+  }
+
+  specification = {
```
vmrvictor, line 13: I know that in projects we have by now only get projects but I don't think we should call the tool at the same way that one endpoint, so it can cause later confusion with next ones... f.e: I do not expect having toos for create-test-case but to have test case where we could do different actions...

src/tests/unit/zephyr/common/api-client.test.ts
assisblf: That's true! Addressed on bf6c52c4ebceaac4ddcd794b5b6f640962b3227d

src/zephyr/tool/get-projects.ts
assisblf: The idea is that each separate class will be a different tool. It makes it easier to make a shorter scope.

For example:
* **Get Projects** and **Get Project** are different tools; therefore, they will be different classes

We can also discuss this too

src/zephyr/tool/get-projects.ts
```diff
@@ -0,0 +1,40 @@
...
+    title: "Get Projects",
+    summary: "Get details of projects in Zephyr",
```
vmrvictor, line 15: I see we have different tags that we could add like:
```
  readOnly: true,
    idempotent: true,
```

src/zephyr/tool/get-projects.ts
vmrvictor: I would gather them by entity/domain, here we will end up like having an adapter per endpoint (if we translate to our backend project)

src/zephyr/client.ts
```diff
@@ -0,0 +1,38 @@
...
+    baseUrl: string = "https://api.zephyrscale.smartbear.com/v2",
+  ) {
+    this.apiClient = new ApiClient(bearerToken, baseUrl);
+  }
+
+  registerTools(
+    register: RegisterToolsFunction,
+    _getInput: GetInputFunction,
+  ): void {
+    console.log("[ZephyrClient] Starting tool registration...");
```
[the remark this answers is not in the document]
assisblf: It doesn't look like it.
From what I saw, `console.error` is safe to use, but not the other levels.

Reference [here](https://modelcontextprotocol.io/docs/develop/build-server#best-practices-2:~:text=Best%20Practices,stdout%20by%20default)


I'll try and explore the logging options.

src/zephyr/tool/project/get-projects.ts
```diff
@@ -0,0 +1,40 @@
...
+        name: "maxResults",
+        type: MaxResultsSchema,
+        required: false,
+      },
+      {
+        name: "startAt",
+        type: StartAtSchema,
+        required: false,
+      },
+    ],
```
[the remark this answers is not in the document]
assisblf: 2722a4011298300de54e0eb22b1fe9eb156f6ce4

vmrvictor (approved):

tomlongridge (commented):
LGTM 👍

PR #172, opened 2025-10-06
refactor(bugsnag): put next URL and count into base structure

tomlongridge:
## Goal

Apply the pattern of having a `link` field for the next URL and `X-Total-Count` header for the total count to the base API code for each request, adding it where it's present.

Plus some extra standardisation along the way.

## Changeset

* Moved the logic out of the couple of client functions that used it, and standardised the passing of response data back
* Added stability data to the project object and cache to prevent an additional call to get it for stability purposes
  * Also moved `Project` interface to the `Project.ts` file for a more natural home, removing `ProjectApiView` as unnecessary
* Added `top_only` parameter to `listReleases` and defaulted to `false` so we're not just returning the top ones 
* In the base `request` method, I've renamed and reversed the `paginate` parameter to `fetchAll` which seemed (to me) a better description as the previous "paginate" was a bit ambiguous and indeed was implemented the opposite way to what I expected to be
  * I've made it default to fetching all as the caller has to know to handle the next URL
* Standardised the parameter names, which were a mix of snake and camel, plus the naming to be `nextUrl` throughout.

## Testing

Unit tests plus manual checks that pagination still works as expected.

### DIFF
--- a/src/bugsnag/client.ts
+++ b/src/bugsnag/client.ts
@@ -589,14 +589,12 @@ export class BugsnagClient implements Client {
         // Get the latest event for this error using the events endpoint with filters
         let latestEvent = null;
         try {
-          const eventsResponse = await this.errorsApi.getLatestEventOnProject(
-            project.id,
-            listEventsQueryString,
-          );
-          latestEvent =
-            eventsResponse.body && eventsResponse.body.length > 0
-              ? eventsResponse.body[0]
-              : null;
+          latestEvent = (
+            await this.errorsApi.getLatestEventOnProject(
+              project.id,
+              listEventsQueryString,
+            )
+          ).body;
         } catch (e) {
           console.warn("Failed to fetch latest event:", e);
           // Continue without latest event rather than failing the entire request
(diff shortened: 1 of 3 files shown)


src/bugsnag/client.ts
```diff
@@ -862,23 +847,21 @@ export class BugsnagClient implements Client {
...
 
         const result = {
-          data: errors,
-          count: errors.length,
-          total: totalCount ? parseInt(totalCount, 10) : undefined,
-          next: linkHeader?.match(/<([^>]+)>/)?.[1],
+          errors: errors,
+          next_url: response.nextUrl,
+          page_error_count: errors.length,
+          total_error_count: response.totalCount ?? undefined,
```
sazap10, line 864: is this going to confuse the LLM that there are this many errors (not bugsnag errors!) in the page and total?

src/bugsnag/client/api/Error.ts
```diff
@@ -150,41 +150,25 @@ export class ErrorAPI extends BaseAPI {
...
-      ? `/errors/${errorId}/latest_event?${params}`
-      : `/errors/${errorId}/latest_event`;
-    return (await this.request<ViewLatestEventOnErrorResponse>({
-      method: "GET",
-      url,
-    })) as ApiResponse<ViewLatestEventOnErrorResponse>;
-  }
-
   /**
    * List the Events on a Project
```
sazap10, line 154: the comment is wrong now

sazap10 (approved):

PR #177, opened 2025-10-09
refactor: move auth to within clients and create common client init

sazap10:
## Goal

Moves auth requirements to within clients and make client initialization into a separate module.

## Design

Moving auth to within the client makes it easier to identify what configuration is required for each client. Also this is a precursor to adding support for http/sse.

## Changeset

- Adds client registry to handle client initialization
- Moves env config to within clients

## Testing

- Local testing to ensure clients still start correctly


sazap10:
closing in favour of the https://github.com/SmartBear/smartbear-mcp/pull/196 pr which adds hosted server support and builds on top of this. Most the comments have been addressed via a refactor

PR #223, opened 2025-11-06
Initial Collaborator client implementation

mohammedzahid-chaudhary:
## Goal

Initial collaborator client for MCP.

## Changeset

Added new tools for Review management and Remote system configuration management.

## Testing

Tested using MCP inspector.

### DIFF
--- a/src/collaborator/client.ts
+++ b/src/collaborator/client.ts
@@ -63,29 +63,17 @@ export class CollaboratorClient implements Client {
         }),
       },
       async (args, _extra) => {
-        try {
-          const { reviewId } = args;
-          const commands = [
-            {
-              command: "ReviewService.findReviewById",
-              args: { reviewId },
-            },
-          ];
-          const result = await this.call(commands);
-          return {
-            content: [{ type: "text", text: JSON.stringify(result, null, 2) }],
-          };
-        } catch (err) {
-          return {
-            content: [
-              {
-                type: "text",
-                text: `Error: ${err instanceof Error ? err.message : String(err)}`,
-              },
-            ],
-            isError: true,
-          };
-        }
+        const { reviewId } = args;
+        const commands = [
+          {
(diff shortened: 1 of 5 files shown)


src/collaborator/client.ts
```diff
@@ -0,0 +1,553 @@
...
+        } catch (err) {
+          return {
+            content: [
+              {
+                type: "text",
+                text: `Error: ${err instanceof Error ? err.message : String(err)}`,
+              },
+            ],
+            isError: true,
+          };
```
tomlongridge: You don't have to do this – an thrown error will be caught in the `server.ts` and a formatted response returned. This also has the benefit of logging the error to BugSnag.

src/index.ts
```diff
@@ -92,9 +96,20 @@ async function main() {
...
+        collaboratorLoginTicket,
+      ),
+    );
+    client_defined = true;
+  }
+
   if (!client_defined) {
     console.error(
-      "Please set one of REFLECT_API_TOKEN, BUGSNAG_AUTH_TOKEN, API_HUB_API_KEY, QMETRY_API_KEY, ZEPHYR_API_TOKEN, or PACT_BROKER_BASE_URL / (and relevant Pact auth) environment variables",
+      "Please set one of REFLECT_API_TOKEN, BUGSNAG_AUTH_TOKEN, API_HUB_API_KEY, QMETRY_API_KEY, ZEPHYR_API_TOKEN, COLLABORATOR_BASE_URL(and relevant collab environment variables) or PACT_BROKER_BASE_URL / (and relevant Pact auth) environment variables",
```
tomlongridge: I think this error message is now stretched beyond its initial purpose. I suggest we just replace with:

```suggestion
      "No SmartBear products were configured. Please provide at least one of the required configuration options.",
```

tomlongridge (commented):
Please also update the top-level README.md with a reference to these tools.

PR #229, opened 2025-11-13
[bugsnag] add "get current project" tool

tomlongridge:
## Goal

The current design is not great at handling project IDs when no API key is provided in configuration. It often finds the List Projects tool but only gives the user a few to choose from if they haven't been specific about which project they are referring to.

## Design

The PR adds a few conveniences that should improve this:

* Added a "Get Current Project" tool that informs the user of the current/default project or throws an error if one is not set, with hints about calling "List Projects" and/or locating the API key from the user's code
* "List Projects" now takes an optional `apiKey` parameter as a convenience for getting a project by API key
* If the user only has access to 1 project in the org, this is set to the current project
* When a tool is called with a project ID and there is no configured current project, that project is set as the current project

## Changeset

* There was also a bug/limitation with the "List Project Event Filters" as it was assuming the project ID of the current projects – I've normalised this to take an optional project ID like the other tools and also updated the cache so that it's not just caching the current project's filters.

## Testing

Updated tests and manual confirmation.


src/bugsnag/client.ts
```diff
@@ -745,22 +772,20 @@ export class BugsnagClient implements Client {
...
+          content: [{ type: "text", text: JSON.stringify(eventFilters) }],
         };
       },
     );
 
     const updateErrorInputSchema = z.object({
       projectId: this.projectApiKey
-        ? toolInputParameters.projectId.optional()
+        ? toolInputParameters.projectId
         : toolInputParameters.projectId,
```
gingerbenw: unnecessary ternary operation

```suggestion
projectId: toolInputParameters.projectId,
```

gingerbenw (approved):
one suggestion, otherwise looks good!

PR #252, opened 2025-11-20
Add network endpoint grouping tools

gingerbenw:
## Goal

This PR adds functionality for managing network endpoint grouping rules in BugSnag projects, enabling users to configure how network spans are grouped for performance monitoring.

## Changeset

- Adds two new tools: "Get Network Endpoint Groupings" and "Set Network Endpoint Groupings"
- Implements API methods to retrieve and update network grouping rulesets
- Includes comprehensive documentation with examples for OpenAPI path templating syntax

### DIFF
--- a/src/tests/unit/bugsnag/client.test.ts
+++ b/src/tests/unit/bugsnag/client.test.ts
@@ -32,6 +32,7 @@ const mockProjectAPI = {
   getReleaseGroup: vi.fn(),
   listBuildsInRelease: vi.fn(),
   getProjectNetworkGroupingRuleset: vi.fn(),
+  updateProjectNetworkGroupingRuleset: vi.fn(),
   getProjectSpanGroup: vi.fn(),
   getProjectSpanGroupDistribution: vi.fn(),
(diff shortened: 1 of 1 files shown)


src/bugsnag/client.ts
```diff
@@ -1527,6 +1527,145 @@ export class BugsnagClient implements Client {
...
+          "Understand how network requests are being grouped in performance monitoring",
+          "Check grouping patterns before making updates",
+        ],
+        inputSchema: getNetworkGroupingInputSchema,
+        examples: [
+          {
+            description: "Get network grouping rules for a project",
+            parameters: {},
+            expectedOutput:
+              "JSON object with projectId and array of endpoint URL patterns",
```
tomlongridge: Shall we just return the array of URLs? The containing object seems a bit unnecessary

src/tests/unit/bugsnag/client.test.ts
tomlongridge: We probably want a couple of unit tests to cover the new tools?

tomlongridge (approved):

PR #255, opened 2025-11-25
fix(bugsnag): update generated API class

tomlongridge:
## Goal

Updates `src/bugsnag/client/api/api.ts` with latest changes from the Data Access API.

## Changeset

Added extra instructions to the top of the file to describe the process of updating the client and removing the unwanted content.

## Testing

Unit and manual tests.


tomlongridge:
Superseded by #292

PR #282, opened 2025-12-12
Enable BugSnag error and performance on by default

tomlongridge:
## Goal

Tracks errors and performance of the MCP server.

## Design

Uses the OTel Node SDK configured for BugSnag Performance monitoring, with attributes for the transport mode and OS.

The `MCP_SERVER_BUGSNAG_API_KEY` can still be used to set the destination of the data, but if not specified it will go to SmartBear's instance. The variable can be set to an empty value to disable all instrumentation.

## Changeset

I've moved the environment variable reading to a `config.ts` file so it can be shared.

Release stages are used to have BugSnag enabled for end-users when using a published SDK and not in development. This is updated by a new `build:production` script that's triggered by the `prepare` phase and the Docker build.

## Testing

Confirmed the data appears (or not) when configured.


package.json
```diff
@@ -32,11 +32,13 @@
   },
   "scripts": {
     "build": "tsc && shx chmod +x dist/*.js",
+    "build:production": "npm run build && npm run build:set-release-stage",
+    "build:set-release-stage": "sed -i.bak 's/MCP_SERVER_RELEASE_STAGE = \"development\"/MCP_SERVER_RELEASE_STAGE = \"production\"/g' dist/common/config.js && rm dist/common/config.js.bak",
```
sazap10: this isnt ideal for anyone wanting to run in windows, maybe we could do this with node instead?

Dockerfile
```diff
@@ -8,13 +8,14 @@ COPY package.json package-lock.json tsconfig.json ./
...
 
-RUN npm run build
+RUN npm run build:production
 
 FROM node:22-alpine AS release
 
 COPY --from=builder /app/dist /app/dist
 COPY --from=builder /app/package.json /app/package.json
 COPY --from=builder /app/package-lock.json /app/package-lock.json
+COPY --from=builder /app/scripts/set-release-stage.js /app/scripts/set-release-stage.js
```
sazap10: you also need to add the script in the builder stage, adding something like:

`COPY scripts/set-release-stage.js ./scripts/set-release-stage.js`
 above in the builder stage should fix this copy as well

PR #306, opened 2026-01-21
fix: add pollyfills for elicitation and sampling

kevinrvaz:
## Goal

This PR adds pollyfills for sampling and elicitations during the stdio & http transport init to enable use of pactflow tools with advanced features in tools which dont support these features. Other features like prompts, resources don't need to be disabled as they are handled by capability negotiation see - https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#capability-negotiation.

## Testing

Tested the changes in vscode copilot, cursor, windsurf and claude code.

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -26,12 +26,18 @@ import {
 import Bugsnag from "../common/bugsnag";
 import { CacheService } from "./cache";
 import { MCP_SERVER_NAME, MCP_SERVER_VERSION } from "./info";
+import {
+  executeElicitationOrPolyfill,
+  isElicitationPolyfillResult,
+} from "./pollyfills";
 import { ToolError } from "./tools";
 import type { Client, ToolParams } from "./types";
 import { unwrapZodType } from "./zod-utils";
 
 export class SmartBearMcpServer extends McpServer {
   private cache: CacheService;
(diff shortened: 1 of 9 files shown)


src/common/server.ts
```diff
@@ -123,10 +167,19 @@ export class SmartBearMcpServer extends McpServer {
             try {
               return await cb(url, variables, extra);
             } catch (e) {
-              Bugsnag.notify(e as unknown as Error, (event) => {
-                event.addMetadata("app", { resource: name, url: url });
-                event.unhandled = true;
-              });
+              Bugsnag.notify(
```
kevinrvaz: this was a linting issue mentioned by biome, no changes here

src/common/server.ts
```diff
@@ -95,18 +117,40 @@ export class SmartBearMcpServer extends McpServer {
                   ],
                 };
               } else {
-                Bugsnag.notify(e as unknown as Error, (event) => {
-                  event.addMetadata("app", { tool: toolName });
-                  event.unhandled = true;
-                });
+                Bugsnag.notify(
```
kevinrvaz: this was a linting issue mentioned by biome, no changes here

JP-Ellis (approved):

PR #322, opened 2026-02-10
POC MCP App

domarmstrong:
## Goal

Investigate MCP Apps and how they could be integrated

## Design

Rather than the examples of MCP apps which have an html entry for each tool and then bundle it's resources in to a single file. I wanted to proof out being able to set up an easier to expand approach which requires less set up for each tool.

* Make it easy to add ui to each client
* Make it easy to add new tools 

### Example MCP app

* Starts when the bugsnag list-projects tool is called
* Shows an interactive list of project names
* Clicking a project will perform a further tool call to get the errors for the project

#### Example when run in vs code

I found running the app in vs code painful as it seems to need completely quitting to update the mcp server.

https://github.com/user-attachments/assets/10a6a933-47b5-49c2-87d0-6728a05e0c6a

#### Example when server is run with UI_DEV

Hot reloading works completely making development not a complete nighmare

https://github.com/user-attachments/assets/32c8e136-a551-48ba-8923-1614a15de828

## Changeset

This is a POC and not expected to merge in the current state.

* Issues around correct handling of CORs. Currently had to allow origin `*` to work with the basic-host for testing. So that needs more investigation
* explicit localhost urls should be probably be handled better
* Static /asset serving is rough

## Testing

Tested via vscode copilot + mcp basic-host example.


src/bugsnag/ui/ListProjects.tsx
```diff
@@ -0,0 +1,108 @@
...
+          )
+          .map((p) => (
+            <ProjectListItem key={p.id} project={p} />
+          ))}
+      </ul>
+    </div>
+  );
+}
+
+interface ErrorResult {
```
domarmstrong: tool results should probably all be exported in some common types for sharing

src/tests/unit/bugsnag/client.test.ts
```diff
@@ -771,10 +774,6 @@ describe("BugsnagClient", () => {
   });
 
   describe("API methods", async () => {
-    beforeEach(async () => {
-      client = await createConfiguredClient("test-token", "test-project-key");
```
domarmstrong: this setup is currently bad. if you don't run all tests then later tests will fail because they are relying on setup from previous tests to exist and not have been cleaned up.

PR #330, opened 2026-02-13
feat: OAuth support

sazap10:
## Goal

To introduce support for Dynamic Request-Level Authentication (OAuth) across the SmartBear MCP Server. Previously, the server required static, server-level credentials to be provided via environment variables at startup, making it unsuitable for multi-tenant environments or usage behind an API gateway where the active user context changes per request. This change allows the server to extract authentication tokens dynamically from incoming HTTP request headers and implements RFC 9293 OAuth discovery.

## Design

To achieve request-level authentication without polluting the `@modelcontextprotocol/sdk` signatures, the solution leverages Node's `AsyncLocalStorage` to propagate the HTTP request headers deep into the tool execution layer. 

1. **`AsyncLocalStorage` for Request Context:** A new `requestContextStorage` wraps the execution of incoming MCP stream/HTTP requests. Tool implementations can call `getRequestHeader()` deep within their business logic to pull context tokens securely and dynamically.
2. **OAuth Discovery Flow:** The HTTP transport layer now exposes the `/.well-known/oauth-protected-resource` endpoint. When a client connects without credentials, the server responds with a `401 Unauthorized` and includes a `WWW-Authenticate` header, pointing the client to the Authorization Server to begin the OAuth flow.
3. **Lazy API Client Initialization:** Client `ConfigurationSchema` definitions (e.g., BugSnag, Qmetry, Zephyr, Swagger, etc.) have been updated to mark authentication fields as `.optional()`. The `isConfigured()` checks have been relaxed to allow the tools to successfully mount to the server at startup, deferring the credential requirement until the actual tool is invoked (at which point it checks the dynamic request context or falls back to the static config).

## Changeset

- **Core/Transport:**
  - Added `src/common/request-context.ts` implementing `AsyncLocalStorage` to store `IncomingMessage` headers.
  - Updated `src/common/transport-http.ts` to wrap `handleRequest` and `handlePostMessage` in `requestContextStorage.run()`.
  - Added the `/.well-known/oauth-protected-resource` metadata endpoint.
  - Added the `WWW-Authenticate` header to `401` initialization failures.
  - Added the `OAUTH_AUTHORIZATION_SERVER_URL` config variable.
- **Client Implementations:**
  - Updated `BugsnagClient` to use a dynamic `apiKey` resolver function instead of a hardcoded string.
  - Updated `ReflectClient`, `ZephyrClient`, `SwaggerClient`, `PactflowClient`, `QmetryClient`, and `CollaboratorClient` to extract tokens via `getRequestHeader()`.
  - Modified `ConfigurationSchema` in all clients to mark `api_key` / `token` / `username` / `login_ticket` as `.optional()`.
  - Updated `isConfigured()` logic across all clients to ensure they mount when static credentials are omitted.

## Testing

- Validated that `npm run build`, `npm run format`, `npm run lint`, and `npm run test` pass successfully.
- Added specific unit tests verifying that API configurations fetch tokens dynamically via the function resolver.
- Verified that tools successfully mount to the `clientRegistry` even when no authentication environment variables are supplied at startup.

## How to test with local development
- Run mcp server with:
```
PORT=4000 MCP_TRANSPORT=http MCP_CLIENTS=bugsnag ALLOWED_ORIGINS=http://localhost:6274,http://localhost:8080,http://localhost:8081 OAUTH_AUTHORIZATION_SERVER_URL=http://localhost:7070 node dist/index.js
```

- Run oauth server
- Run bugsnag with oauth support


vmrvictor (approved):
for zephyr it is working

kordian-smartbear (approved):

KrzysztofNocen (approved):

src/common/transport-http.ts
```diff
@@ -20,6 +31,7 @@ export async function runHttpMode() {
   const allowedOrigins = process.env.ALLOWED_ORIGINS?.split(",") || [
     "http://localhost:3000",
   ];
+  const baseUrlOverride = process.env.BASE_URL; // Allow explicit override if headers are unreliable
```
tomlongridge: Consider adding to `getBaseUrl` so the logic is all in one place

src/common/transport-http.ts
```diff
@@ -60,7 +73,10 @@ export async function runHttpMode() {
         return;
       }
 
-      const url = new URL(req.url || "/", `http://${req.headers.host}`);
+      // Determine the public URL of this server
+      // Use env override if set, otherwise detect from request headers
+      const baseUrl = baseUrlOverride || getBaseUrl(req);
+      const url = new URL(req.url || "/", baseUrl);
```
tomlongridge, line 83: Consider also adding to `getBaseUrl`

src/common/transport-http.ts
```diff
@@ -451,18 +495,52 @@ async function newServer(
...
+        // Client doesn't support dynamic auth — auth was provided at config time
+        if (!client.getAuthToken) return true;
+        // Client supports dynamic auth — check if a token is available
+        return client.getAuthToken() !== null;
+      }),
+    );
+
+    if (!hasAuth) {
+      throw new Error(
+        "No clients have valid authentication credentials. Please authenticate via OAuth or provide auth headers.",
```
tomlongridge:
```suggestion
        "No clients have valid authentication credentials. Please authenticate via OAuth or provide alternative auth headers (e.g. API key or personal auth token).",
```

src/common/transport-http.ts
```diff
@@ -421,10 +461,14 @@ async function handleLegacyMessageRequest(
     try {
       const parsedBody = JSON.parse(body);
       // Route message to the SSE transport for processing
-      await (session.transport as SSEServerTransport).handlePostMessage(
-        req,
-        res,
-        parsedBody,
+      await requestContextStorage.run(
```
tomlongridge: Is it worth abstracting the `run` method so that it takes a request? Then it's not up to the caller to create the store with headers in.

src/bugsnag/client.ts
```diff
@@ -69,7 +70,12 @@ interface StabilityData {
 }
 
 const ConfigurationSchema = z.object({
-  auth_token: z.string().describe("BugSnag personal authentication token"),
+  auth_token: z
+    .string()
+    .describe(
+      "BugSnag authentication token (personal access token or OAuth token)",
```
tomlongridge: Is this true? The config value is just for using with a PAT isn't it?

src/common/transport-http.ts
```diff
@@ -451,18 +495,52 @@ async function newServer(
...
+          const headerName = getHeaderName(client, key);
+          // Check both original case and lower-case headers for compatibility
+          // (HTTP headers are case-insensitive, but Node.js lowercases them)
+          const value =
+            req.headers[headerName] || req.headers[headerName.toLowerCase()];
+          if (typeof value === "string") {
+            return value;
+          }
+
+          return null;
```
tomlongridge: How about using `getEnvVarName` from the stdio server as a fallback? i.e. so that we use the environment variable name if it's present and not in the header.

That should avoid the extra logic around endpoint I think?

src/common/transport-http.ts
sazap10: moved to `getBaseUrl`

src/common/transport-http.ts
sazap10: moved to `getBaseUrl`

src/common/transport-http.ts
sazap10: updated

src/common/transport-http.ts
sazap10: changed to add a new function `withRequestContext` which takes a request

src/bugsnag/client.ts
sazap10: nope, updated that description

src/common/transport-http.ts
sazap10: changed to fallback to `getEnvVarName`

src/bugsnag/client.ts
```diff
@@ -113,13 +115,79 @@ export class BugsnagClient implements Client {
     config: z.infer<typeof ConfigurationSchema>,
   ): Promise<void> {
     this.cache = server.getCache();
+
+    // Allow reading endpoint from environment variable as a fallback
+    // This is useful for On-Premise installations where the endpoint is fixed
+    const endpoint = config.endpoint || process.env.BUGSNAG_ENDPOINT;
```
tomlongridge: No longer required I think?

src/collaborator/client.ts
```diff
@@ -49,11 +50,25 @@ export class CollaboratorClient implements Client {
...
+    const contextLogin = getRequestHeader("Collaborator-Login");
+    const contextTicket = getRequestHeader("Collaborator-Ticket");
+
+    if (contextLogin) {
+      login = Array.isArray(contextLogin) ? contextLogin[0] : contextLogin;
+    }
+    if (contextTicket) {
+      ticket = Array.isArray(contextTicket) ? contextTicket[0] : contextTicket;
+    }
+
```
tomlongridge, line 66: Is this missing a fallback to the configured username/ticket?

src/common/request-context.ts
```diff
@@ -0,0 +1,39 @@
+import { AsyncLocalStorage } from "node:async_hooks";
+import type { IncomingMessage } from "node:http";
+
+// Define the shape of our request context
```
tomlongridge: I feel like here it would be appropriate to explain the purpose:

e.g.

```suggestion
// Storage for pre-request data that can be retrieved from a tool to prevent caching as part of the server instance in a session.
// For example, the auth token.
```

src/bugsnag/client.ts
sazap10: yup, removed

src/pactflow/client.ts
```diff
@@ -240,13 +232,58 @@ export class PactflowClient implements Client {
...
+      };
+    }
+    return undefined;
   }
 
   async getResult<T>(resultUrl: string): Promise<T> {
     const response = await fetch(resultUrl, {
       method: "GET",
-      headers: this.headers,
+      headers: this.requestHeaders,
```
tomlongridge, line 286: Not sure if I'm missing something, but will that method be executed or should it be:
```suggestion
      headers: this.requestHeaders(),
```

tomlongridge (approved):

PR #338, opened 2026-02-19
[BugSnag] PLAT-14898: Added the ability to add comments to BugSnag errors

jasonMcCullough:
## Goal

Adds the ability for comments to be added to errors in BugSnag.

## Design

This coincides with how comments are normally created via the dashboard. Only with this, we're exposing it for being called via MCP instead.

## Changeset

- createCommentOnError exposes the `/comments` endpoint and manages the necessary query parameter and header logic.
- `createCommentOnErrorInputSchema` contains the Zod logic for defining how interactions should be structured when making calls to the `/comments `endpoint.

## Testing

image
image

### DIFF
--- a/src/bugsnag/client/api/Error.ts
+++ b/src/bugsnag/client/api/Error.ts
@@ -211,4 +211,31 @@ export class ErrorAPI extends BaseAPI {
       false, // Paginate results
     );
   }
+
+  /**
+   * Create a comment on an Error
+   * POST /projects/{project_id}/errors/{error_id}/comments
+   */
+  async createCommentOnError(
+    projectId: string,
+    errorId: string,
+    message: string,
+  ): Promise<ApiResponse<any>> {
+    const url = new URL(
+      `/projects/${projectId}/errors/${errorId}/comments`,
+      this.configuration.basePath,
+    );
+
+    const options = {
+      method: "POST",
+      headers: {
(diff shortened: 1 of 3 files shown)


src/bugsnag/client/api/Error.ts
```diff
@@ -211,4 +211,31 @@ export class ErrorAPI extends BaseAPI {
...
+    const options = {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "X-Bugsnag-Api": "true",
+        Authorization: this.configuration.apiKey,
+      },
+      body: JSON.stringify({ message }),
+    };
+
```
sazap10: we generate the API code from the openAPI spec which should generate the creator for this request. https://github.com/SmartBear/smartbear-mcp/blob/5d39bac4f9d547a1d8f7ec7b88f8f24ffffab48b/src/bugsnag/client/api/api.ts#L13 is how we do that. We should use something like that for this endpoint

jasonMcCullough: Ah wait - I should clarify this endpoint isn't in our Data Access API so I probably can't finish this to the linked standard before I go 😢 Tom was in touch with the rails team to try and flesh something out for the `/comments` endpoint to be made officially part of our public APIs though

PR #377, opened 2026-03-17
Oauth server scale

vmrvictor:
## Goal

Zephyr remote server


sazap10:
@vmrvictor i presume we can close this now

PR #388, opened 2026-03-24
feat: convert bugsnag tools to typesafetools

sazap10:
## Goal

Move bugsnag tools to use TypeSafeTool class from https://github.com/SmartBear/smartbear-mcp/pull/368. No functional changes


domarmstrong (commented):
Well, I wasn't quite ready to put it up for review as I was planning on revisiting the API for the tools. But it's probably worth going ahead with and I can revisit after.

PR #406, opened 2026-04-02
feat: add List Events on an Error tool

tomlongridge:
## Goal

Adds the ability to query events from an error – for example to do an event-by-event comparison using an LLM.

## Design

Copied the pattern of other similar tools.

## Testing

Unit testing of the basics and real dashboard testing of the tool.

Example prompt:

> give the latest 5 events for error that occurred most recently


sazap10 (approved):
We should add something to the changelog, but otherwise seems good to me

sazap10 (approved):

PR #433, opened 2026-04-24
[Bugsnag] Add ability to assign/unassign error to collaborator

DariaKunoichi:
## Goal
Add ability to assign user to errors.

## Changeset
User can be assigned using ID or name.
Added new tool for listing collaborators in a project so that AI can query them if only user's name is provided to extract ID. 

## Testing
Tested manually with a dashboard project.

### DIFF
--- a/src/bugsnag/client/api/Organization.ts
+++ b/src/bugsnag/client/api/Organization.ts
@@ -1,0 +1,47 @@
+import { type ApiResponse, BaseAPI } from "./base";
+import {
+  type Collaborator,
+  OrganizationsApiFetchParamCreator,
+} from "./index";
+
+export class OrganizationAPI extends BaseAPI {
+    static collaboratorFields: (keyof Collaborator)[] = [
+    "id",
+    "name",
+    "email",
+    "two_factor_enabled",
+    "two_factor_enabled_on",
+    "recovery_codes_remaining",
+    "password_updated_on",
+    "show_time_in_utc",
+    "heroku",
+    "created_at",
+    "is_admin",
+    "pending_invitation",
+    "last_request_at",
+    "project_ids",
+    "paid_for",
+    "project_roles",
+    "managed_by_smartbear_id",
+  ];
+
+    /**
+     * List the collaborators in a Project
+     * GET /projects/{project_id}/collaborators
+     * @param projectId The project ID
+     * @returns A promise that resolves to the list of collaborators
+     */
+    async listProjectCollaborators(
+      projectId: string,
+    ): Promise<ApiResponse<Collaborator[]>> {
+      const localVarFetchArgs = OrganizationsApiFetchParamCreator(
+        this.configuration,
+      ).listProjectCollaborators(projectId);
+      return await this.requestArray<Collaborator>(
+        localVarFetchArgs.url,
+        localVarFetchArgs.options,
+        true,
+        OrganizationAPI.collaboratorFields,
+      );
+    }
+}
--- a/src/bugsnag/client/api/Organization.ts
+++ b/src/bugsnag/client/api/Organization.ts
@@ -1,47 +1,26 @@
 import { type ApiResponse, BaseAPI } from "./base";
-import {
-  type Collaborator,
-  OrganizationsApiFetchParamCreator,
-} from "./index";
+import { type Collaborator, OrganizationsApiFetchParamCreator } from "./index";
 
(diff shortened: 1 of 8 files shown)


src/bugsnag/client/api/api.ts
```diff
@@ -638,6 +638,128 @@ interface ErrorReopenRules {
...
+ * @export
+ * @interface ProjectRoles
+ */
+export interface ProjectRoles {
+    /**
+     *
+     * @type {string}
+     * @memberof ProjectRoles
+     */
+    _515fb9337c1074f6fd000001: string;
```
gingerbenw: @SmartBear/bugsnag-dev Can I just confirm with the pipeline team that this type is valid? Would a generic  `[key: string]` be better? What exactly is this id?

gingerbenw (commented):
seems fine to me, one small comment I think might be worth checking

src/bugsnag/client/api/api.ts
sazap10: well its mapping project ids to a role for the user

gingerbenw: so it shouldn't be this fixed value?

sazap10: Correct

gingerbenw:
```suggestion
    [id: string]: string;
```

src/bugsnag/tool/error/update-error.ts
```diff
@@ -171,6 +187,46 @@ export class UpdateError extends Tool<BugsnagClient> {
...
+      {
+        description: "Unassign an error from a collaborator",
+        parameters: {
+          errorId: "6863e2af8c857c0a5023b411",
+          operation: "assign",
+          assigned_collaborator_id: null,
+        },
+        expectedOutput:
+          "Success response indicating the error was unassigned from the collaborator",
+      },
```
tomlongridge: In order to prevent too many examples for this tool - I suggest just keeping user assign and team unassign, to cover options

src/bugsnag/tool/error/update-error.ts
```diff
@@ -182,6 +238,8 @@ export class UpdateError extends Tool<BugsnagClient> {
       "For 'n_occurrences_in_m_hours' reopen rules, specify both 'occurrences' and 'hours' parameters",
       "For 'n_additional_occurrences' reopen rules, specify 'additionalOccurrences' parameter",
       "Snoozing temporarily silences an error until the specified reopen condition is met",
+      "'assigned_collaborator_id' and 'assigned_team_id' are mutually exclusive when using 'assign' operation - only one can be provided to assign to either a collaborator or a team",
```
tomlongridge:
```suggestion
      "'assigned_collaborator_id' and 'assigned_team_id' are mutually exclusive - either assign to a collaborator (user) or a team. Use `null` to unassign.",
```

src/bugsnag/tool/project/list-collaborators.ts
```diff
@@ -0,0 +1,69 @@
...
+          projectId: "6863e2af8c857c0a5023b411",
+          collaboratorNameOrId: "John Doe",
+        },
+        expectedOutput:
+          "JSON object with collaborator details including name, email, and role",
+      },
+    ],
+    hints: [
+      "Collaborator name or ID is used to filter the list of collaborators for a project. If a name is provided, it will be matched against collaborator names to find the corresponding ID.",
+    ],
```
tomlongridge: This doesn't appear to be implemented - i.e. match by name. I think that's correct - it should just return a list - we just need to update the tool description.

src/bugsnag/client/api/Organization.ts
```diff
@@ -0,0 +1,47 @@
...
+    "show_time_in_utc",
+    "heroku",
+    "created_at",
+    "is_admin",
+    "pending_invitation",
+    "last_request_at",
+    "project_ids",
+    "paid_for",
+    "project_roles",
+    "managed_by_smartbear_id",
```
tomlongridge: Let's not actively expose this – I realise it's available in the API, but I can't see a strong use case for having it in our model.

src/bugsnag/tool/error/update-error.ts
DariaKunoichi: I'll probably remove team mentions as I did not add a tool to list teams.

PR #435, opened 2026-04-28
[Reflect] RF-4313: Updated get-screenshot tool with jpeg format support and added oauth for some mcp tools.

dhavalsinh-chavda:
## Goal

Add OAuth support to some of the Reflect MCP tools so that requests authenticated via Authorization: Bearer token are handled correctly alongside the existing API key auth (X-API-KEY/Reflect-Api-Token). The list-segments endpoint also requires a different base URL when accessed via OAuth, and get-screenshot needed a format option to support JPEG in addition to PNG.
 
## Design

Auth detection is centralised in two new methods on ReflectClient:                                                             
  
 - isOAuthRequest() — checks whether the incoming request carries an Authorization: Bearer header, respecting the existing header priority (Reflect-Api-Token and X-API-KEY take precedence over Authorization).
 - getAuthHeader() — returns the correct auth header object (Authorization: Bearer <token> for OAuth, X-API-KEY: <token> for API key) based on the result of isOAuthRequest().                                                                       
  
WebSocketManager was updated to accept a Record<string, string> auth header instead of a raw API key string, so it works transparently for both auth types.                 
          
 For list-segments, the base URL is selected at call time based on isOAuthRequest() — OAuth requests route to app.reflect.run/api/mcp, API key requests continue to use api.reflect.run/v1. 
          
 The format parameter on get-screenshot defaults to png if not supplied, keeping the change fully backwards compatible.
## Changeset

- client.ts — added isOAuthRequest() and getAuthHeader(); updated getHeaders() to delegate to getAuthHeader()                                        
 - websocket-manager.ts — constructor now accepts authHeader: Record<string, string> instead of apiKey: string
 - connect-to-session.ts — passes getAuthHeader() to WebSocketManager instead of getApiToken()                                                
 - list-segments.ts — selects API base URL based on auth type; uses getHeaders() for all request headers                                           
 - get-screenshot.ts — added optional format parameter (png | jpeg, defaults to png); mimeType in the response reflects the chosen format                          
 - types/mcp.ts — added format field to MCPGetScreenshotMessage                                                               
 - constants.ts — added WEB_APP_HOSTNAME constant for the OAuth base URL
 
## Testing

- [x] Verified list_sgements, get-screenshot and connect to session with oAuth token
- [x] Verified list_sgements, get-screenshot and connect to session with API Key

### DIFF
--- a/src/reflect/client.ts
+++ b/src/reflect/client.ts
@@ -88,14 +88,36 @@ export class ReflectClient implements Client {
     return this.getAuthToken() || "";
   }
 
-  getHeaders(): Record<string, string> {
+  isOAuthRequest(): boolean {
+    if (
+      getRequestHeader("Reflect-Api-Token") ||
+      getRequestHeader("X-API-KEY")
+    ) {
+      return false;
+    }
+    const authHeader = getRequestHeader("Authorization");
+    return (
+      !!authHeader &&
+      (Array.isArray(authHeader) ? authHeader[0] : authHeader).startsWith(
+        "Bearer ",
+      )
+    );
+  }
+
+  getAuthHeader(): Record<string, string> {
     const token = this.getAuthToken();
     if (!token) {
       throw new Error("Reflect API token not found");
     }
+    if (this.isOAuthRequest()) {
+      return { Authorization: `Bearer ${token}` };
+    }
+    return { [API_KEY_HEADER]: token };
+  }
 
+  getHeaders(): Record<string, string> {
     return {
-      [API_KEY_HEADER]: token,
+      ...this.getAuthHeader(),
       "Content-Type": "application/json",
       "User-Agent": `${MCP_SERVER_NAME}/${MCP_SERVER_VERSION}`,
     };
--- a/src/reflect/client.ts
+++ b/src/reflect/client.ts
@@ -10,7 +10,11 @@ import type {
   RegisterPromptFunction,
   RegisterToolsFunction,
 } from "../common/types";
-import { API_KEY_HEADER } from "./config/constants";
+import {
+  API_KEY_HEADER,
+  AUTHORIZATION_HEADER,
+  REFLECT_API_TOKEN_HEADER,
+} from "./config/constants";
 import { SapTest } from "./prompt/sap-test";
 import { AddPromptStep } from "./tool/recording/add-prompt-step";
 import { AddSegment } from "./tool/recording/add-segment";
@@ -60,9 +64,9 @@ export class ReflectClient implements Client {
   getAuthToken(): string | null {
     // 1. Try request context
     const contextHeader =
-      getRequestHeader("Reflect-Api-Token") ||
-      getRequestHeader("X-API-KEY") ||
-      getRequestHeader("Authorization");
+      getRequestHeader(REFLECT_API_TOKEN_HEADER) ||
+      getRequestHeader(API_KEY_HEADER) ||
+      getRequestHeader(AUTHORIZATION_HEADER);
 
     if (contextHeader) {
       let token = Array.isArray(contextHeader)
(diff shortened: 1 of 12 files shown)


src/tests/unit/reflect/tool/recording/get-screenshot.test.ts
```diff
@@ -84,4 +86,26 @@ describe("GetScreenshot", () => {
...
+
+  it("should default to png format when format is not specified", async () => {
+    await instance.handle({ sessionId: "sess-1" }, {});
+
+    expect(mockWsManager.sendMcpMessage).toHaveBeenCalledWith(
+      expect.objectContaining({ type: "mcp:get-screenshot", format: "png" }),
+    );
+  });
+
+  it("should use jpeg format when format is jpeg", async () => {
```
ihasfrozen, line 98: [optional] Could this also include a test for `png` being specified?

src/reflect/client.ts
```diff
@@ -88,14 +88,36 @@ export class ReflectClient implements Client {
     return this.getAuthToken() || "";
   }
 
-  getHeaders(): Record<string, string> {
+  isOAuthRequest(): boolean {
+    if (
+      getRequestHeader("Reflect-Api-Token") ||
+      getRequestHeader("X-API-KEY")
```
ihasfrozen: Can these strings go into constants for re-use between here and `getAuthToken`?

src/reflect/client.ts
```diff
@@ -88,14 +88,36 @@ export class ReflectClient implements Client {
     return this.getAuthToken() || "";
   }
```
ihasfrozen: Is this method still used?

src/tests/unit/reflect/tool/tests/list-segments.test.ts
ihasfrozen, line 73: Can these assertions also be included in your new test?

ihasfrozen (commented):
What is expected to happen for tool calls that don't currently have OAuth support, such as `list-suites`? As-written, I expect those calls will 404, but should there be some more graceful error handling in that case?

Could `registerTools` filter down to only the options that have OAuth support when it detects that OAuth is being used?

src/tests/unit/reflect/client.test.ts
```diff
@@ -46,4 +47,76 @@ describe("ReflectClient", () => {
   it("should have correct config prefix", () => {
     expect(client.configPrefix).toBe("Reflect");
   });
+
+  describe("isOAuthRequest", () => {
+    it("should return true when Authorization header starts with Bearer", () => {
```
mahavir1166, line 47: Could you add one small test for a lowercase `authorization` header with a lowercase `bearer` scheme? That would cover the HTTP header casing we usually see from Node, and it would protect the case-insensitive OAuth detection behavior if we update it.

src/reflect/client.ts
```diff
@@ -88,14 +88,36 @@ export class ReflectClient implements Client {
...
+      getRequestHeader("X-API-KEY")
+    ) {
+      return false;
+    }
+    const authHeader = getRequestHeader("Authorization");
+    return (
+      !!authHeader &&
+      (Array.isArray(authHeader) ? authHeader[0] : authHeader).startsWith(
+        "Bearer ",
+      )
```
mahavir1166: Could you make this `Bearer` check case-insensitive? For example, `authorization: bearer <token>` would not currently be detected as OAuth. Normalizing the auth scheme before checking it would make this a bit more forgiving and closer to how HTTP auth schemes are usually handled.

dhavalsinh-chavda:
> What is expected to happen for tool calls that don't currently have OAuth support, such as `list-suites`? As-written, I expect those calls will 404, but should there be some more graceful error handling in that case?
> 
> Could `registerTools` filter down to only the options that have OAuth support when it detects that OAuth is being used?

Currently it is returning unauthorised, but I updated code to filter down only OAuth supported tools.

src/reflect/client.ts
```diff
@@ -173,6 +199,10 @@ export class ReflectClient implements Client {
       new AddSegment(this),
     ];
 
+    const tools = this.isOAuthRequest()
+      ? oAuthAndAPISupportedTools
+      : [...oAuthAndAPISupportedTools, ...apiOnlyTools];
```
ihasfrozen, line 204: Just double-checking - were you able to test this locally to confirm it works correctly?

ihasfrozen (approved):

mahavir1166 (approved):
🏁

src/reflect/client.ts
dhavalsinh-chavda: Yes @ihasfrozen, I tested that and with API Key its shows 14 tools and with oAuth token it shows 6 tools.
Screenshot 2026-04-29 at 11 17 43 AM
Screenshot 2026-04-29 at 11 18 40 AM

PR #442, opened 2026-05-05
TM4J-17231 Update test case and cycle fix

kordian-smartbear:


vmrvictor (approved):

src/zephyr/common/rest-api-schemas.ts
```diff
@@ -1145,29 +1141,19 @@ export const UpdateTestCaseBody = zod
...
-        ),
-    }),
+      .number()
+      .min(1)
+      .nullable()
+      .optional()
+      .describe(
+        "The ID of the folder, to remove folder set it's value to null",
+      ),
+    owner: zod
```
juro994, line 1151: Maybe a stupid question. But how would the MCP know that it has to fill in `owner.accountId` just based on this schema? Wouldn't it try to send JSON where the `owner` itself is a number? Wouldn't that fail on the side of our REST API?

src/zephyr/tool/test-case/update-test-case.ts
```diff
@@ -77,14 +77,44 @@ export class UpdateTestCase extends Tool<ZephyrClient> {
...
+
+    const nullValuesObject: Record<string, any> = {};
+
+    if (rawUpdates.folder) {
+      //do nothing when null or undefined
+      nullValuesObject.folder = { id: rawUpdates.folder };
+    }
+    if (rawUpdates.owner) {
+      //do nothing when null or undefined
+      nullValuesObject.owner = { id: rawUpdates.owner };
```
vmrvictor: this should be accountId

vmrvictor (changes requested):

src/zephyr/tool/test-case/update-test-case.ts
vmrvictor:
```suggestion
      nullValuesObject.owner = { accountId: rawUpdates.owner };
```

src/zephyr/common/rest-api-schemas.ts
vmrvictor: https://github.com/SmartBear/smartbear-mcp/pull/442#discussion_r3187395320 you are right we are setting id and not accountId in the code

juro994: Ok, just adding a note here for visibility. This is handled manually in the merging logic in the tool. There the `owner: 123` is transformed to `owner: { accountId: 123 }`. So it doesn't really matter that schema doesn't match our REST API.

adrienave: I think it's covered by the changes on lines 105 to 107 of `update-test-case.ts` tool implementation.

So the short answer: call to the API itself is done by our tool implementation after transformation of the inputs received from MCP, AI agent doesn't need to know anything about our API.

EDIT: keeping my comment for reference but Juraj found the answer by himself 😛

vmrvictor (approved):

vmrvictor (approved):

PR #446, opened 2026-05-05
refactor: update all tools to use inputSchema rather than parameters

tomlongridge:
## Goal

For a while we have had both `parameters` and `inputSchema` in the `ToolParams` interface – the latter allows the input schema to be defined once and used in both the implementation and the definition: keeping them in sync and type-safe.

## Changeset

Remove `ToolParams.parameters` and updated all uses to use `inputSchema` with the equivalent Zod definition.

## Testing

Removed some assertions on the parameters - these seemed unneccessary as there are already tools covering this for the common code (`src/tests/unit/common/server.test.ts`).

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -108,8 +108,10 @@ export class SmartBearMcpServer extends McpServer {
           {
             title: toolTitle,
             description: this.getDescription(params),
-            inputSchema: this.schemaToRawShape(params.inputSchema),
-            outputSchema: this.getOutputSchema(params),
+            inputSchema: params.inputSchema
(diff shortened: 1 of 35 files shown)


tmcneal (approved):
LGTM. I also pulled down the branch and was able to run various Reflect tools successfully 👍

sazap10 (approved):

tomlongridge:
This is now rebased on top of the snapshot tests and fixes. Small re-review, if you wouldn't mind?

tmcneal (approved):

sazap10 (approved):

PR #447, opened 2026-05-05
feat: enforce CHANGELOG.md updates when src/ is modified

assisblf:
## Goal

Ensure contributors always update `CHANGELOG.md` when they make changes to files under `src/`, so the changelog stays accurate and up to date.

## Design

Added a `changelog-check` CI job to the existing Test Suite workflow (`test.yml`). The job:
- Only performs the check on `pull_request` events (steps are gated with `if: github.event_name == 'pull_request'`), so it passes trivially on direct pushes to `main` without being skipped — which keeps the `complete` gate job happy.
- Uses `git diff` against the PR base branch to detect whether any `src/` files were changed without a corresponding `CHANGELOG.md` update, and fails with a clear `::error::` annotation if so.

## Changeset

- `.github/workflows/test.yml`: added `changelog-check` job and added it to the `complete` job's `needs` list.

## Testing

- [x] Create a dummy change to confirm the new check will fail if `CHANGELOG.md` is not changed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - build
+      - changelog-check
       - coverage
       - format
       - lint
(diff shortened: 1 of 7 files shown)


AdamSmartBear (approved):

tomlongridge (approved):
Approving to test the check...

assisblf:
Taking a screenshot just to show what it'd look like if the changelog check fails.
image

tomlongridge (approved):

PR #450, opened 2026-05-06
test: add snapshot tests to detect changes in tool definitions

tomlongridge:
## Goal

The tool definition output is pretty key to the success of the server and liable to changes that are hard to spot during development.

## Design

Added a snapshot test to catch changes to the output – so make them pass, the updated version must be checked in (run `vitest -u` to regenerate it) and reviewed.

## Changeset

New `src/tests/unit/common/registered-tools.test.ts` using vitest functionality.

## Testing

Automated testing.


sazap10 (approved):
My browser is really lagging on the files changes screen, but it looks plausible to me

tomlongridge:
> My browser is really lagging on the files changes screen, but it looks plausible to me

Yeah - it won't mean much here, but the diff on the next PR if it changes will be informative.

PR #455, opened 2026-05-07
feat(common): graceful shutdown on SIGTERM/SIGINT + split health/readiness endpoints

sazap10:
## Goal

Adds graceful shutdown so that on SIGTERM the process:

1. Stops advertising itself as ready (so the LB stops sending new sessions).
2. Drains active sessions, running each client's `cleanupSession` hook.
3. Exits 0 on a clean drain or 1 if a configurable deadline is exceeded.

It also splits the existing `/health` endpoint into two probes: liveness and readiness.

## Design

**ShutdownManager** (`src/common/shutdown.ts`) is a small lifecycle module with SIGTERM/SIGINT handling. Subsystems register named cleanup callbacks; on signal they run in **LIFO order** with a deadline (default 25s, overridable via `MCP_SHUTDOWN_TIMEOUT_MS`).

**Probe split.** 
- `/health` is now liveness-only 
- `/ready` is readiness and returns 503 once `isDraining()` is true. 

**HTTP transport drain** (`drainHttpTransport`) sequences:

1. `httpServer.close()` to stop accepting new TCP connections.
2. `httpServer.closeIdleConnections()` to release keep-alive sockets with no in-flight work.
3. `transport.close()` on every entry in the `transports` map. This ends SSE streams and triggers the existing `transport.onclose` → `server.cleanupSession(sessionId)` chain, which is where Reflect's
per-session WebSocket teardown actually runs.
4. `httpServer.closeAllConnections()` as a backstop so `httpServer.close()` resolves even if a client refuses to drop its socket.

**Stdio transport** registers a much smaller handler that just calls `transport.close()`. Process-wide teardown for stdio (e.g. closing all Reflect WebSockets) would need a new `Client.cleanupAll()` interface method, out of scope here.

## Changeset

- `src/common/shutdown.ts` - added shutdown manager class 
- `src/common/transport-http.ts`- added `handleHealthRequest`, `handleReadyRequest`, `drainHttpTransport` methods
- `src/common/transport-stdio.ts` - added shutdown handler
- added tests for new http endpoints and shutdown handling
- new env var: `MCP_SHUTDOWN_TIMEOUT_MS` (default `25000`).

## Testing
No change to mcp tools functionality

**Manual test**: started in HTTP mode, hit both probes, sent SIGTERM:

```
=== /health (should be 200 ok) ===
HTTP/1.1 200 OK
Cache-Control: no-store
{"status":"ok","timestamp":"..."}

=== /ready (should be 200 ready) ===
HTTP/1.1 200 OK
Cache-Control: no-store
{"status":"ready","timestamp":"..."}

=== Sending SIGTERM ===
[MCP][shutdown] Received SIGTERM, starting drain
[MCP][shutdown] Draining 1 handler(s), deadline 5000ms
[MCP][shutdown] Draining HTTP transport (0 active session(s))
[MCP][shutdown] HTTP transport drained
[MCP][shutdown] Drained cleanly in 1ms

=== Process exited with code 0 ===
```

### DIFF
--- a/src/common/shutdown.ts
+++ b/src/common/shutdown.ts
@@ -1,0 +1,253 @@
+/**
+ * Graceful shutdown coordination.
+ *
+ * Owns SIGTERM / SIGINT handling and runs registered cleanup callbacks in
+ * LIFO order with a deadline. Designed so each transport / subsystem can
+ * register its own teardown without knowing about the others.
+ *
+ * Lifecycle:
+ *   idle      -> initial state
+ *   draining  -> first signal received; isDraining() returns true so
+ *                readiness probes can flip to 503 and tools can refuse new
+ *                work if they choose to. Handlers are running.
+ *   complete  -> handlers finished or deadline hit; process is about to exit.
+ *
+ * Double-signal escalation: a second SIGTERM/SIGINT during drain forces an
+ * immediate exit(1) without waiting for outstanding handlers.
+ *
+ * Exit codes:
+ *   0  clean drain
+ *   1  deadline exceeded, or forced exit via second signal
+ */
+
+export type ShutdownHandler = () => Promise<void> | void;
+
+interface RegisteredHandler {
+  name: string;
+  fn: ShutdownHandler;
+}
+
+export interface ShutdownManagerOptions {
+  /** Signals to listen for. Defaults to ['SIGTERM', 'SIGINT']. */
+  signals?: NodeJS.Signals[];
+  /** Drain deadline in milliseconds. Defaults to MCP_SHUTDOWN_TIMEOUT_MS or 25000. */
+  timeoutMs?: number;
+  /** Process to attach handlers to. Injectable for tests. */
+  process?: NodeJS.Process;
+  /** Exit function. Injectable for tests. */
+  exit?: (code: number) => void;
+  /** Logger. Injectable for tests. */
+  logger?: Pick<Console, "log" | "warn" | "error">;
+}
+
+export interface DrainResult {
+  status: "clean" | "deadline";
+  durationMs: number;
+  /** Names of handlers that did not complete before the deadline. */
+  remainingHandlers: string[];
+}
+
+const DEFAULT_TIMEOUT_MS = 25_000;
+
+function defaultTimeout(): number {
+  const fromEnv = process.env.MCP_SHUTDOWN_TIMEOUT_MS;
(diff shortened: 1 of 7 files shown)


src/common/transport-http.ts
```diff
@@ -157,6 +202,62 @@ export async function runHttpMode() {
...
+ *      This does NOT close existing keep-alive / SSE connections.
+ *   2. Close idle keep-alive connections immediately.
+ *   3. Close every active transport, which fires transport.onclose →
+ *      cleanupSession(sessionId) → per-client cleanupSession (e.g. Reflect
+ *      WebSockets).
+ *   4. Wait for httpServer.close() to fully resolve, then force-close any
+ *      remaining keep-alive connections as a backstop.
+ */
+export async function drainHttpTransport(
+  httpServer: import("node:http").Server,
```
tomlongridge: This feels a bit unconventional way of importing?

src/common/shutdown.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    if (winner === "clean") {
+      this.logger.log(`[MCP][shutdown] Drained cleanly in ${durationMs}ms`);
+      return { status: "clean", durationMs, remainingHandlers: [] };
+    }
+
+    this.logger.error(
+      `[MCP][shutdown] Deadline exceeded after ${durationMs}ms, ${remainingHandlers.length} handler(s) outstanding: ${remainingHandlers.join(", ") || "(none)"}`,
+    );
+    return { status: "deadline", durationMs, remainingHandlers };
```
tomlongridge: Maybe cleaner?

```suggestion
    if (winner === "clean") {
      this.logger.log(`[MCP][shutdown] Drained cleanly in ${durationMs}ms`);
    } else {
      this.logger.error(
        `[MCP][shutdown] Deadline exceeded after ${durationMs}ms, ${remainingHandlers.length} handler(s) outstanding: ${remainingHandlers.join(", ") || "(none)"}`,
      );
    }
    return { status: winner, durationMs, remainingHandlers };
```

sazap10: yeh, updated

src/common/transport-http.ts
sazap10: yeh, moved import at the top

tomlongridge (approved):

tomlongridge (approved):

PR #457, opened 2026-05-07
Adds manifest and script for packaging as an MCP bundle

tomlongridge:
## Goal

Adds support for the [MCPB](https://github.com/modelcontextprotocol/mcpb) format.

Run `npm run pack:mcpb` to generate the `smartbear.mcpb` file.

## Design

To avoid proliferation of config files at the top level, the script compiles the bundle in a temporary build directory. This also means the minimal files go in without complicated ignore files.

(I think this could be a direction we take the other packaging scripts in due course.)

## Changeset

* New manifest.json in the [appropriate format](https://github.com/modelcontextprotocol/mcpb/blob/main/MANIFEST.md)
* Added a new script (pack-mcpb.js) to copy the files to a build directory and generate the bundle

## Testing

Extensive testing through Claude code extensions.


src/common/transport-http.ts
sazap10, line 1: we shouldnt have this change in this PR

package.json
```diff
@@ -44,7 +44,8 @@
     "test:coverage:ci": "vitest --coverage --reporter=verbose",
     "test:run": "vitest run",
     "coverage:check": "vitest --coverage --reporter=verbose --config vitest.config.coverage.ts",
-    "bump": "node scripts/bump.js"
+    "bump": "node scripts/bump.js",
+    "pack:mcpb": "npm run build && node scripts/pack-mcpb.js"
```
sazap10: thoughts on using pre script for this https://docs.npmjs.com/cli/v8/using-npm/scripts?v=true#pre--post-scripts:


```suggestion
    "prepack:mcpb": "npm run build",
    "pack:mcpb": "node scripts/pack-mcpb.js"
```

packaging/manifest.json
```diff
@@ -0,0 +1,220 @@
...
+        "PACT_BROKER_USERNAME": "${user_config.PACT_BROKER_USERNAME}",
+        "QMETRY_API_KEY": "${user_config.QMETRY_API_KEY}",
+        "QMETRY_BASE_URL": "${user_config.QMETRY_BASE_URL}",
+        "REFLECT_API_TOKEN": "${user_config.REFLECT_API_TOKEN}",
+        "SWAGGER_API_KEY": "${user_config.SWAGGER_API_KEY}",
+        "SWAGGER_PORTAL_BASE_PATH": "${user_config.SWAGGER_PORTAL_BASE_PATH}",
+        "SWAGGER_REGISTRY_BASE_PATH": "${user_config.SWAGGER_REGISTRY_BASE_PATH}",
+        "SWAGGER_UI_BASE_PATH": "${user_config.SWAGGER_UI_BASE_PATH}",
+        "ZEPHYR_API_TOKEN": "${user_config.ZEPHYR_API_TOKEN}",
+        "ZEPHYR_BASE_URL": "${user_config.ZEPHYR_BASE_URL}"
```
sazap10, line 62: would be nice to generate this from the code but maybe a future enhancement?

.gitignore
```diff
@@ -3,3 +3,4 @@ dist
 .vscode/
 .idea/
 coverage
+build/
```
sazap10, line 6: also add the generated bundle?

```suggestion
build/
smartbear.mcpb
```

packaging/manifest.json
tomlongridge: I have a prototype :) I might get chance to push it today – it will also generate the tool definitions and we can include in the other manifests and docs.

package.json
tomlongridge: It feels like there should be a better syntax, but yes, I agree.

In fact, if I called the tasks `pare` then it would presumably run the existing `prepare` task ;-)

tomlongridge: In fact, that syntax interferes a little with the built-in `pack` script. I've now made this `postpack` so that it runs as part of `npm pack` and uses `prepare` on the way.

src/common/transport-http.ts
tomlongridge: I've rebased on main now that #453 is merged.

package.json
sazap10: seems good to me

sazap10 (approved):

PR #458, opened 2026-05-08
Standardise prompt and resource registration

tomlongridge:
## Goal

We have defined types for Tool parameters but not Resources or Prompts, so some naming inconsistencies have crept in, and may continue, between products. I'm working on a script to automatically generate the documentation by extracting tools, resources and prompts, so fixing this will mean that it can parse them in a consistent way: allowing them to be split between the products.

## Design

This PR applies the same pattern to registering of prompts and resources as for tools – using the common server class to prefix them with the product name to disambiguate.

## Testing

Amended unit test and confirmed resources and prompts work in VSCode.

I've also added the resources and prompts to the snapshot testing to guard against unintended changes in future.

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -41,20 +41,11 @@ export class SmartBearMcpServer extends McpServer {
         version: MCP_SERVER_VERSION,
       },
       {
-        instructions: `When creating or editing a Reflect test using a connected recording session, follow these guidelines:
-
-1. After connecting to a session, get the list of segments for the session's platform type so you know what actions could be added via segments vs needing to create new steps. Do not list tests, only list segments.
-2. Before performing an action, take a screenshot to understand the current state of the application.
-3. Each add_prompt_step request should perform a single action or assertion. Do not combine multiple actions or assertions into a single step.
-4. Only perform one action at a time unless you're sure the action won't move the application to a different screen. For example, you can send multiple add_prompt_step requests to fill out individual form fields if those fields are visible on the current screen.
-5. Check the list of existing Segments to see if a Segment exists that achieves a similar goal to what you're trying to do next. If so, add the segment instead of creating new steps.
-6. If a step fails, use delete_previous_step to remove it and try a different approach.
-7. After completing a task, if the task required multiple prompt steps, add a final prompt step that validates the current state of the page based on what you see on the screen. In your validation, do not reference information that can change from run to run.`,
         capabilities: {
-          resources: { listChanged: true }, // Server supports dynamic resource lists
           tools: { listChanged: true }, // Server supports dynamic tool lists
+          resources: { listChanged: true }, // Server supports dynamic resource lists
+          prompts: { listChanged: true }, // Server supports sending prompts to Host
           logging: {}, // Server supports logging messages
-          prompts: {}, // Server supports sending prompts to Host
         },
       },
     );
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -95,8 +95,8 @@ export class SmartBearMcpServer extends McpServer {
     this.clients.push(client);
     await client.registerTools(
       (params, cb) => {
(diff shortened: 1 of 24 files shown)


src/common/server.ts
```diff
@@ -41,20 +41,11 @@ export class SmartBearMcpServer extends McpServer {
         version: MCP_SERVER_VERSION,
       },
       {
-        instructions: `When creating or editing a Reflect test using a connected recording session, follow these guidelines:
```
tomlongridge, line 44: As discussed elsewhere, this prompt is now in the Reflect tool and so not required at server level anymore.

src/common/server.ts
```diff
@@ -172,14 +163,17 @@ export class SmartBearMcpServer extends McpServer {
     );
 
     if (client.registerResources) {
-      client.registerResources((name, path, cb) => {
-        const url = `${client.toolPrefix}://${name}/${path}`;
+      await client.registerResources((params, cb) => {
+        const url = `${client.toolPrefix}://${params.name}/${params.path}`;
```
sazap10: given that we are using getToolName for the resource name which handles spaces in the name, we should do the same for the url


```suggestion
        const slug = params.name.replace(/\s+/g, "_").toLowerCase();
        const url = `${client.toolPrefix}://${slug}/${params.path}`;
```

src/common/server.ts
```diff
@@ -205,12 +202,28 @@ export class SmartBearMcpServer extends McpServer {
...
-        return super.registerPrompt(name, config, cb);
+      await client.registerPrompts((params, cb) => {
+        return super.registerPrompt(
+          this.getToolName(client, params.title),
+          {
+            title: this.getToolTitle(client, params.title),
+            description: params.description,
+            argsSchema: this.schemaToRawShape(params.argsSchema) || {},
+          },
+          (args: any, extra: any) => cb(args, extra),
```
sazap10: should this have a bugsnag error handler?

sazap10: also was there a reason to wrap the `cb` in a function rather than just passing it directly to `registerPrompt`?

src/common/server.ts
```diff
@@ -205,12 +202,28 @@ export class SmartBearMcpServer extends McpServer {
...
+        );
       });
     }
   }
 
+  private getToolTitle(client: Client, title: string): string {
+    return `${client.name}: ${title}`;
+  }
+
+  private getToolName(client: Client, title: string): string {
```
sazap10: given that we are using this for more than just tools now we should rename to something more generic like `getCapabilityName`?

sazap10 (commented):
A couple minor comments

src/common/server.ts
tomlongridge: It is now we have a Bugsnag handler

src/common/server.ts
```diff
@@ -172,31 +157,29 @@ export class SmartBearMcpServer extends McpServer {
...
+        const url = `${client.capabilityPrefix}://${slug}/${params.path}`;
         return super.registerResource(
-          name,
+          this.getCapabilityName(client, params.name),
           new ResourceTemplate(url, {
             list: undefined,
           }),
-          {},
+          {
+            title: this.getCapabilityTitle(client, params.name),
```
sazap10: do we want to add a title to the resource params as well like we do for tools as we will currently get something like `BugSnag: event` for the bugsnag event resource (lower case 'e') title

src/pactflow/client.ts
```diff
@@ -2421,9 +2421,16 @@ export class PactflowClient implements Client {
    *
    * @param register - The function used to register prompts.
    */
-  registerPrompts(register: RegisterPromptFunction): void {
+  async registerPrompts(register: RegisterPromptFunction): Promise<void> {
     PROMPTS.forEach((prompt) => {
```
sazap10: not necessarily related to this change but we should use `for ... of` here rather than `forEach`. `for...of` works with async functions as well

sazap10 (approved):

PR #462, opened 2026-05-12
feat: update all action versions and improve testing

sazap10:
Bumps action versions and improves code coverage action


.github/workflows/docs-as-code.yaml
```diff
@@ -8,8 +8,6 @@ on:
       - 'docs/**'
   release:
     types: [published]
-    paths:
```
sazap10, line 11: this is not valid for release event

.github/workflows/docs-as-code.yaml
```diff
@@ -75,11 +73,10 @@ jobs:
     runs-on: ubuntu-latest
     environment: ${{ github.event_name == 'release' && 'Production' || github.event.inputs.environment || 'Staging' }}
     needs: [spell-check, validate-manifests]
-    if: github.event.inputs.skip_api_linting == 'true' || (needs.spell-check.result == 'success' && needs['validate-manifests'].result == 'success')
```
sazap10, line 78: the input `skip_api_linting` doesnt exist (i guess it was added for testing) and the other conditions are covered by the `needs` enforcement

tomlongridge (approved):

PR #470, opened 2026-05-14
Fixes required for MCPB publishing

tomlongridge:
## Goal

Some smaller fixes found during MCPB testing.

## Changeset

* Set the author.url in manifest to GitHub (as per Anthropic guidelines)
* Validate that tool names are <= 64 characters (https://github.com/anthropics/claude-code/issues/34960)
* Reduced the offending tool names

## Testing

Added a runtime exception when registering tools to ensure they cannot go above 64 again. This triggers unit test failures in the snapshot test if violated, so we will have future CI errors if it happens again.


sazap10 (approved):
Do we want to add a github workflow to error on tool name length?

tomlongridge:
> Do we want to add a github workflow to error on tool name length?

Apologies - I forgot to add a note to the Testing section above: we're protected from regression by a runtime exception that will be caught by the snapshot unit tests if any tool goes above the limit.

sazap10:
> > Do we want to add a github workflow to error on tool name length?
> 
> Apologies - I forgot to add a note to the Testing section above: we're protected from regression by a runtime exception that will be caught by the snapshot unit tests if any tool goes above the limit.

ah nice.

sazap10 (approved):

PR #473, opened 2026-05-15
feat: create GitHub release with MCPB attached from tag

tomlongridge:
## Goal

It would be useful to attach `.mcpb` files to the GitHub release for reference. To facilitate this I've amended the publish job to automatically create (or update) the release with the artifact.

I've also taken the opportunity to script the adding of the changelog to the release notes.

## Design

Added `publish_mcpb` parameter for manual dispatch to run later if required.

Also added `tag` input on which to run the script - this ensures that all the build and publish steps act on the same commit.

## Testing

Cloned the repo and ran the scripts with a new tag and updating existing releases.

This hasn't been tested with the rest of the publishing, but this will need to happen on next release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,8 +160,9 @@ Please follow these steps to create a new release:
     - Get approval and merge the PR into `main`
 
 2. **Release**
-    - Create and push a tag for the new release (e.g., `git tag v1.2.3` and `git push origin v1.2.3`)
-    - This will trigger the publish workflow, which will publish the new version to external sites (e.g. npm) and create the GitHub release
+    - Create a new release on GitHub (or push a new tag) with the version, including the `v` prefix (e.g. `v1.2.3`)
+      - This will trigger the publish workflow, which will publish the new version to external sites (e.g. npm) and update the GitHub release with the changelog details and artifacts
+    - Await/request approval of the publish workflow
 
 3. **Post-Release**
     - Check the publish action to make sure all deployments succeeded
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -148,23 +148,20 @@ A release will be carried out by a member of the SmartBear team when a set of re
...
-   ```
 1. **Create PR to main**
-    - Push the changes to GitHub and create a Pull Request from your release branch into `main`
+    - Create a Pull Request from your release branch into `main`
+    - Get approval and merge the PR into `main`
 
 2. **Release**
-    - Create a new Github release with the version you decided on earlier
-    - Changes will be automatically deployed once the Github release is created
+    - Create and push a tag for the new release (e.g., `git tag v1.2.3` and `git push origin v1.2.3`)
```
sazap10: the release will create the tag so i dont see a reason to update this

tomlongridge: The publish script is now _triggered_ by the tag - so the process to kick it off is to push it. After which the release is generated.

sazap10: hmm, could we not just attach the mcpb to the release rather than getting publish to create the release? i like the fact you have to create the github release with the release notes rather than just pushing a tag

tomlongridge: Understood - the job will still fire from a tag, but the tag creation can be done as part of creating a release in GitHub. The script supports both.

I've updated the instructions.

sazap10 (approved):

PR #479, opened 2026-05-19
[BearQ] Add BearQ MCP

s-leirbag:
## Goal

Add BearQ MCP support

## Design

I followed the patterns of other projects.

I broke down the "create task" endpoint into multiple tools to more easily surface the different usages to the agent.

## Changeset

Added new BearQ project with new client/tools/docs.

## Testing

Tested locally, verified all tools work as expected

Note: Our task streaming public endpoint is not live just yet but I built for it and tested locally. It should be live soon. *It is now live


s-leirbag:
The code coverage job may be failing because I'm on a fork

src/bearq/client.ts
```diff
@@ -0,0 +1,105 @@
...
+  api_token: z.string().describe("BearQ workspace API token (Bearer)."),
+  api_base_url: z
+    .string()
+    .optional()
+    .describe(
+      "Override the BearQ public API base URL. Defaults to https://api.bearq.smartbear.com",
+    ),
+});
+
+export class BearqClient implements Client {
```
fitzn: nit: I think this would be `BearQClient` to be consistent with the stylized product name, BearQ.

s-leirbag: Fixed

s-leirbag:
Opened an equivalent PR from the main repo here: https://github.com/SmartBear/smartbear-mcp/pull/485

PR #487, opened 2026-05-21
refactor: split authentication from configuration

tomlongridge:
## Goal

With the evolution of the MCP server from stdio to http and then the addition of OAuth support, the original design for "configuration" of each of the clients has been creaking somewhat as the authentication fields aren't always provided upfront.

This means we've added some workarounds to circumvent the Zod-based validation as "required" fields aren't always required. It also means we can't do a Zod `parse` of the config, which applies things like default values, so it can be misleading when you define the configuration schema.

## Design

This PR splits out the authentication fields from configuration explicitly. A client is "configured" once when the server is created using the defined configuration schema, but it can also ask for authentication data via the `server` object when a tool is executed using a new `getEnv` function passed in through the transport layer. This function is responsible for reading the data from the appropriate place (`process.env` or the request) and with the correct prefix for the client.

In other words, all the clients now configure themselves with a function that will be called at tool execution time to get the appropriate authentication data.

### `authenticationFields`

Each client has now had the authentication fields moved from the Zod-defined configuration input (`client.config`) to a new `client.authenticationFields` to define the requirements separately. 

### `client.getAuthToken` -> `client.hasAuth`

In the `client` type, the `getAuthToken` function is replaced by `hasAuth`. This works alongside `isConfigured` for the client registry to be able to test the state of the client. For example, if none of the configured cilents "have authentication data" then it can return a 401 to start the OAuth flow.

### `GetEnvFn`

A type is defined for `GetEnvFn` to help implement it in the two transport layers and associated tests. It is used when applying the configuration values to the schema as well as passed to the server so it can be exposed to each client when required.

The server provides a layer of validation to ensure that the field requested by the client is defined in the `authenticationFields` to prevent drift.

The function takes an optional client – if provided then it adds the defined prefix. The http implementation first checks the query string, then the headers and finally environment variables; each in the expected format. This makes it consistent where any data will come from.

### Removal of `request-context`

Because we create a closure of the environment in the transport layer in the `GetEnvFn`, we no longer need the `request-context` functionality. This also makes things a little cleaner as the clients do not need to be aware of where they are running.

## Testing

Amended unit tests accordingly.

I've also tested each setup (stdio with config, stdio with no config, http and http with OAuth) using BugSnag to check that it behaves as it did before.


src/common/transport-http.ts
```diff
@@ -651,73 +646,97 @@ export async function newServer(
   req: IncomingMessage,
   res: ServerResponse,
 ): Promise<SmartBearMcpServer | null> {
+  const configFn: GetEnvFn = (key, client) => {
```
sazap10: seems like we are doing similar things here and in getConfigValue function, perhaps they could be combined into one function or have a separate helper that are used by both of these?

tomlongridge: It is used in both – see L284 and 288 – one passed to the server for use in the tool; the other passed to the client registry for configuration.

sazap10: yeh i mean combining the lookup into one function, see https://github.com/SmartBear/smartbear-mcp/pull/504

sazap10 (commented):
just needs the changelog changes, otherwise looks good to me

sazap10 (approved):

PR #492, opened 2026-05-26
fix(pactflow): remove .email() Zod validator from admin user tool schemas

### LINKED ISSUE #491: z.string().email() generates JSON Schema with regex lookahead, rejected by OpenAI gpt-5.5
### Describe the bug

`z.string().email()` in Zod v4 generates a JSON Schema `pattern` field containing negative lookahead assertions:

```
^(?!\.)(?!.*\.\.)([A-Za-z0-9_'+\-\.]*)[A-Za-z0-9_+-]@...
```

OpenAI's gpt-5.5 API validates tool input schemas strictly and rejects any schema containing regex lookaround. This causes every response to fail with the above schema in scope.

The affected tools are `bugsnag_contract-testing_admin_create_user`, `bugsnag_contract-testing_admin_update_user`, and the `listUsers` `q` filter,  all in `src/pactflow/client/base.ts`.

### Steps to reproduce

1. Configure `@smartbear/mcp` as an MCP server with an OpenAI gpt-5.5 backend
2. Send any message

### Environment

* IDE: OpenCode 1.15.10
* LLM: OpenAI gpt-5.5

<details><summary>Error messages:</summary>

```
AI_APICallError: Invalid JSON schema: regex lookaround is not supported. Found at $.properties.email.pattern.
```

</details>

**Suggested fix:** Replace `z.string().email()` with `z.string()` at those call sites. Since these are MCP tool input parameters (not runtime validators), the LLM only needs to know the field is a string; the email format constraint isn't needed here. Alternatively, swap the regex for Zod v4's `regexes.html5Email`, which is lookahead-free.

EmCousin:
## Goal

Remove the `.email()` Zod validator from PactFlow admin user tool schemas. Zod v4's `.email()` generates a JSON Schema `pattern` field containing negative lookahead assertions (`(?!\.)`, `(?!.*\.\.)`), which strict JSON Schema validators reject.

Fixes #491

## Design

These schemas are MCP tool input definitions; they describe parameters to the LLM, not runtime validators. The email format constraint adds no value here and the generated pattern breaks compatibility with validators that don't support regex lookaround (notably OpenAI gpt-5.5, which rejects the entire tool list when any schema contains a lookahead pattern).

Replacing `z.string().email()` with `z.string()` is the minimal correct fix: the `describe()` annotation still communicates intent to the LLM.

## Changeset

- `src/pactflow/client/base.ts`: remove `.email()` from `CreateAdminUserSchema`, `UpdateAdminUserSchema`, and `InviteUsersSchema`
- `CHANGELOG.md`: add entry under Unreleased

## Testing

Existing test suite passes. The affected schemas have no tests that assert on the `.email()` constraint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- [PactFlow] Remove `.email()` Zod validator from PactFlow admin user tool schemas — the generated JSON Schema pattern used regex lookahead which is rejected by strict JSON Schema validators (e.g. OpenAI gpt-5.5) [#491](https://github.com/SmartBear/smartbear-mcp/issues/491)
+
 ## [0.23.0] - 2026-05-22
 
 ### Added
--- a/src/pactflow/client/base.ts
+++ b/src/pactflow/client/base.ts
@@ -662,7 +662,7 @@ export const AdminUserIdSchema = z.object({
 export type AdminUserIdInput = z.infer<typeof AdminUserIdSchema>;
 
 export const CreateAdminUserSchema = z.object({
-  email: z.string().email().describe("Email address of the new user"),
+  email: z.string().describe("Email address of the new user"),
   name: z.string().describe("Display name of the new user"),
   firstName: z.string().optional().describe("First name"),
   lastName: z.string().optional().describe("Last name"),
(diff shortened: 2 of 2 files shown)


kevinrvaz (approved):
LGTM

sazap10 (approved):

EmCousin (commented):
**Note on the approach:** A reviewer might wonder whether `z.email()` (Zod v4's standalone email type) would be a better fix, preserving `"format": "email"` in the generated JSON Schema while dropping the pattern. I tested this via the actual MCP SDK conversion path (`toJsonSchemaCompat`) and both `z.string().email()` and `z.email()` produce identical output, including the lookahead pattern:

```
z.string().email() -> {"type":"string","format":"email","pattern":"^(?!\\.)(?!.*\\.\\.)(...)"}
z.email()          -> {"type":"string","format":"email","pattern":"^(?!\\.)(?!.*\\.\\.)(...)"}
z.string()         -> {"type":"string"}
```

`z.string()` is the only option that avoids the lookahead, so removing `.email()` is not just pragmatic, it's the only clean fix available with the current Zod v4 + MCP SDK stack.

kevinrvaz:
@EmCousin there is a restriction at our end that the pr needs to be raised by one of our internal engineers i've cherry-picked your changes into a new pr #495

kevinrvaz:
changes merged as part of #495 thanks for contributing

PR #495, opened 2026-05-26
fix(pactflow): remove .email() Zod validator from admin user tool sch…

### LINKED ISSUE #491: z.string().email() generates JSON Schema with regex lookahead, rejected by OpenAI gpt-5.5

kevinrvaz:
Zod v4's .email() generates a JSON Schema pattern with negative lookahead assertions, which strict validators (e.g. OpenAI gpt-5.5) reject. These are MCP tool input schemas, not runtime validators, so the email format constraint is unnecessary.

Fixes #491

## Goal

cherry-picking changes from https://github.com/SmartBear/smartbear-mcp/pull/492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- [PactFlow] Remove `.email()` Zod validator from PactFlow admin user tool schemas — the generated JSON Schema pattern used regex lookahead which is rejected by strict JSON Schema validators (e.g. OpenAI gpt-5.5) [#491](https://github.com/SmartBear/smartbear-mcp/issues/491)
+
 ## [0.23.0] - 2026-05-22
 
 ### Added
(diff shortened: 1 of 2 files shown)


Ilia (approved):

EmCousin:
Thanks @kevinrvaz!

PR #499, opened 2026-05-27
feat: add Swagger Functional Testing integration

bartoszkosicki13:
Goal
  
  Extend the existing Swagger integration in the SmartBear MCP Server with Swagger Functional Testing support. Per RF-4662, users can now discover API tests from their Swagger Functional Testing account
  via the MCP Server. Authentication uses a separate SWAGGER_FUNCTIONAL_TESTING_API_TOKEN env var (or X-API-KEY request header for multi-tenant/remote use). FT tools are opt-in: they only appear when
  the token is configured — no separate client registration is needed.

  Design

  Rather than a new standalone client, Functional Testing is added as an optional capability of the existing SwaggerClient. A FunctionalTestingAPI helper class (inside src/swagger/client/) handles token
  resolution and HTTP calls to api.reflect.run/v1/tests. The SwaggerClient conditionally instantiates it during configure() and skips registering FT tools when no token is present. This avoids
  introducing a new client lifecycle and keeps auth token resolution consistent with the existing Swagger pattern (request header → configured token fallback).
  
  Changeset

  - src/swagger/client/functional-testing-api.ts — new FunctionalTestingAPI class; resolves X-API-KEY from request header or configured token, builds headers, calls api.reflect.run/v1/tests
  - src/swagger/client/functional-testing-tools.ts — defines the List Tests tool config (toolset: Functional Testing, handler: listFunctionalTestingTests)
  - src/swagger/client/tools.ts — spreads FUNCTIONAL_TESTING_TOOLS into the master TOOLS array
  - src/swagger/client.ts — adds functional_testing_api_token to ConfigurationSchema; instantiates FunctionalTestingAPI when token is present; adds getFtAuthToken() and listFunctionalTestingTests()
  methods; skips FT toolset in registerTools() when not configured
  - docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md — new doc page describing the list_tests tool and SWAGGER_FUNCTIONAL_TESTING_API_TOKEN requirement
  - docs/products/SmartBear MCP Server/getting-started.md — adds Functional Testing to the product comparison table
  - docs/products/SmartBear MCP Server/local-server.md — adds SWAGGER_FUNCTIONAL_TESTING_API_TOKEN to all config examples (VS Code, Claude Desktop, CLI, MCP Inspector)
  - docs/products/SmartBear MCP Server/manifest.json — registers the new doc page
  - CHANGELOG.md — entry under Unreleased

  Testing
  
  - Unit tests for SwaggerClient FT integration (swagger-client-ft.test.ts): token resolution priority (X-API-KEY header > configured token > null), conditional tool registration (FT tools absent
  without token, present with token), existing Swagger tools unaffected, listFunctionalTestingTests calls https://api.reflect.run/v1/tests with correct headers
  - Unit tests for FunctionalTestingAPI (functional-testing-api.test.ts): getFtHeaders() throws ToolError on missing token, listTests() returns parsed JSON, throws ToolError on non-2xx response
  - All tests via vitest with vitest-fetch-mock


docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
```diff
@@ -0,0 +1,17 @@
+The Swagger Functional Testing client provides tools for discovering API tests. Tools for Swagger Functional Testing require a `FUNCTIONAL_TESTING_API_TOKEN`.
+
+## Available Tools
+
+### Test Discovery
+
+#### `list_tests`
```
KacperSB, line 9: In snapshot you added `functional_testing_list_tests` but here it's `list_tests`
Regardless of what name we agreed on, there should be the same name for both to not create confusion.

Although i still suggest to name it `functional_testing_list_tests` as i don't see any reason not to do it to be honest

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.24.0] - 2026-05-27
+
+### Added
+
+- [Functional Testing] Add Swagger Functional Testing integration with `list_functional_testing_tests` tool for discovering available API tests ([#499](https://github.com/SmartBear/smartbear-mcp/pull/499))
```
KacperSB: And here somehow the name is completely different = )
Please adjust to have 1 tool name everywhere

docs/products/SmartBear MCP Server/local-server.md
```diff
@@ -70,6 +70,10 @@ The SmartBear MCP Server supports multiple SmartBear products, each requiring it
 
   Generate your QTM4J API key by following the instructions [here](https://support.smartbear.com/qmetry-test-management-for-jira-cloud/docs/en/user-guide/qmetry-open-api.html).
 
+- **Swagger Functional Testing**
+
+  Generate an API key from your SmartBear Functional Testing account dashboard at [`app.reflect.run`](https://app.reflect.run/settings/account).
```
KacperSB: Still wrong branding `SmartBear Functional Testing`, should be `Swagger Functional Testing`

docs/products/SmartBear MCP Server/local-server.md
```diff
@@ -121,6 +125,9 @@ export QTM4J_API_KEY="your-qtm4j-api-key"
 # Optional: Set your QTM4J base URL based on your region
 # US (default): https://qtmcloud.qmetry.com Australia: https://syd-qtmcloud.qmetry.com
 export QTM4J_BASE_URL="https://qtmcloud.qmetry.com"
+
+# Required for SmartBear Functional Testing tools
```
KacperSB: and here

docs/products/SmartBear MCP Server/local-server.md
```diff
@@ -259,6 +267,12 @@ Create or edit `.vscode/mcp.json` in your workspace:
       "type": "promptString",
       "description": "US region (default): https://qtmcloud.qmetry.com. Australia region: https://syd-qtmcloud.qmetry.com.",
       "password": false
+    },
+    {
+      "id": "functional_testing_api_token",
+      "type": "promptString",
+      "description": "SmartBear Functional Testing API Token",
```
KacperSB: and here

docs/products/SmartBear MCP Server/local-server.md
```diff
@@ -529,6 +547,12 @@ To run the built server locally in VS Code, add the following to `.vscode/mcp.js
...
```
KacperSB: and here

src/common/register-clients.ts
```diff
@@ -45,3 +46,6 @@ clientRegistry.register(new Qtm4jClient());
 
 // Register Collaborator client
 clientRegistry.register(new CollaboratorClient());
+
+// Register SmartBear Functional Testing client
```
KacperSB: and here

src/functional-testing/client.ts
```diff
@@ -0,0 +1,72 @@
...
+const ConfigurationSchema = z.object({
+  api_token: z.string().describe("Swagger Functional Testing API token"),
+});
+
+export class FunctionalTestingClient implements Client {
+  private _apiToken: string | undefined;
+
+  name = "Swagger Functional Testing";
+  capabilityPrefix = "functional_testing";
+  configPrefix = "Functional-Testing";
```
KacperSB: There is a method called `getQueryStringName()` in repo that generates token/key strings with the client prefix included and formatted. 
It correctly splits strings on "_" and capitalizes, but does nothing for "-", meaning that prefix will go as-is if used.
E.g. for "api_token" it will produce "Functional-TestingApiToken" which is akward. 
Same bug for `PactflowClient` unfortunately...

I suggest to just use CamelCase like many other clients:
```suggestion
  configPrefix = "FunctionalTesting";
```

KacperSB (approved):
Few non-blocking issues, good to go when they are solved or just answered

Kamil-Gumieniuk (approved):
Well done!
I'll be nice to fix naming issues reported by @KacperSB before merging.

src/functional-testing/client.ts
bartoszkosicki13: Thanks for pointing that, however, changing to "FunctionalTesting" has an unintended side effect on the env var name.
  
  getEnvVarName converts the prefix with .toUpperCase().replace(/-/g, "_"), so:
  - "Functional-Testing" → FUNCTIONAL_TESTING_API_TOKEN
  - "FunctionalTesting" → FUNCTIONALTESTING_API_TOKEN
  
The real fix would be in getQueryStringName to also handle - as a word separator, but that would be a breaking change for existing PactFlow users who use query string auth. So I'd suggest keeping "Functional-Testing" as-is for now and tracking the getQueryStringName fix as a separate issue covering both Functional-Testing and Pact-Broker.

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
bartoszkosicki13: The difference between the snapshot and the changelog is intentional- it's the established convention in this project.
  The snapshot always contains the full registered tool name (with capabilityPrefix_ prepended), while the changelog uses the short name without the prefix. This is consistent across all clients:
  - BearQ snapshot: bearq_get_task, bearq_stop_task → changelog: get_task, stop_task
  - Reflect snapshot: reflect_list_segments, reflect_get_screenshot → changelog: list_segments, get_screenshot
  - Swagger snapshot: swagger_create_api_from_prompt → changelog: create_api_from_prompt
  - Functional Testing snapshot: functional_testing_list_tests → changelog: list_tests

KrzysztofNocen (approved):

src/functional-testing/client.ts
KacperSB: True, missed that method
Yeah this code needs improvements, let's not forget about it

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
KacperSB: indeed, all good

KacperSB (approved):
:shipit:

sergioifg94 (approved):

src/functional-testing/tool/list-functional-testing-tests.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  handle: ToolCallback<ZodRawShape> = async (_args) => {
+    const response = await fetch(`https://${API_HOSTNAME}/v1/tests`, {
+      method: "GET",
+      headers: this.client.getHeaders(),
+    });
+
+    if (!response.ok) {
+      throw new ToolError(
+        `Failed to list functional testing tests: ${response.status} ${response.statusText}`,
```
KacperSB:
```suggestion
        `Failed to list Functional Testing tests: ${response.status} ${response.statusText}`,
```

src/tests/unit/functional-testing/tool/list-functional-testing-tests.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    fetchMock.mockResponseOnce(JSON.stringify([]));
+    const result = await instance.handle({}, {} as any);
+    const parsed = JSON.parse((result.content[0] as any).text);
+    expect(parsed).toHaveLength(0);
+  });
+
+  it("should throw ToolError if fetch fails", async () => {
+    fetchMock.mockResponseOnce("Unauthorized", { status: 401 });
+    await expect(instance.handle({}, {} as any)).rejects.toThrow(
+      "Failed to list functional testing tests",
```
KacperSB:
```suggestion
      "Failed to list Functional Testing tests",
```

KacperSB (approved):
Go with it, 2 optional suggestions below

bartoszkosicki13:
Hi, requesting a re-review following recent changes. The implementation was redesigne - SFT is no longer a standalone FunctionalTestingClient but is instead embedded as an optional extension of the existing SwaggerClient. This removes the auth duplication, aligns with how SFT relates to the Swagger product. All tests updated accordingly. Thanks!

src/swagger/client.ts
```diff
@@ -116,6 +134,14 @@ export class SwaggerClient implements Client {
     return this._apiKey || null;
   }
 
+  getFtAuthToken(): string | null {
+    const contextHeader = getRequestHeader("X-API-KEY");
```
KacperSB: You could use the const `FUNCTIONAL_TESTING_API_KEY_HEADER` you exported in `functional-testing-api.ts` instead of hardcoded `"X-API-KEY"`

server.json
```diff
@@ -108,6 +108,12 @@
           "description": "QTM4J Automation API key used exclusively by the automation import tools. Leave empty to disable automation tools. Learn more: https://developer.smartbear.com/smartbear-mcp/docs/qtm4j-integration",
           "isRequired": false,
           "isSecret": true
+        },
```
KacperSB: [optional] I would move it above near the `SWAGGER_API_KEY` so it can be found easier

src/swagger/client/functional-testing-api.ts
```diff
@@ -0,0 +1,38 @@
...
+  ) {}
+
+  getFtHeaders(): Record<string, string> {
+    const token = this.getToken();
+    if (!token) {
+      throw new ToolError("Swagger Functional Testing API token not found");
+    }
+    return {
+      [FUNCTIONAL_TESTING_API_KEY_HEADER]: token,
+      "Content-Type": "application/json",
```
KacperSB, line 19: You using Content-Type `application/json` for GET request which cannot have body 🤔 
Suggest to remove this

src/swagger/client.ts
```diff
@@ -116,6 +134,14 @@ export class SwaggerClient implements Client {
...
+    if (contextHeader) {
+      return Array.isArray(contextHeader) ? contextHeader[0] : contextHeader;
+    }
+    return this._ftApiToken || null;
+  }
```
KacperSB, line 149: This method reads the `X-API-KEY` request header and returns it even when `ftApi` is null (in case when token was not configured, so FT tools were never registered). The return value is unreachable in that state but the method signature makes it look like FT supports a header-only auth flow (like ReflectClient does via `isOAuthRequest()`), but it doesn't: as `listFunctionalTestingTests()` guards on `this.ftApi` and throws `ToolError("Functional Testing API not configured")`.

Same issue is at line 312: the registration guard uses `!this._ftApiToken` instead of `!this.ftApi`. 
Both are in sync today, but `ftApi` is the actual capability state instead of the token.

I would suggest to guard for `!this.ftApi` in both places

KacperSB (commented):
Few suggestions from me

src/swagger/client.ts
bartoszkosicki13: applied

src/swagger/client/functional-testing-api.ts
bartoszkosicki13: Thanks for pointing that. Content-Type is technically unnecessary for GET requests since there's no body. However, getFtHeaders() is a shared method intended for all FT API calls — including POST/PUT methods we plan to add. Keeping it here avoids duplicating it across every method that does have a body.

Kamil-Gumieniuk (approved):
Looks really nice!

Kamil-Gumieniuk:
@bartoszkosicki13 I see that for some reason I don't have access to approve PRs here. Do you know what group should I join?

src/swagger/client/functional-testing-api.ts
KacperSB: ah i see, that's okay, optionally you could add a comment there so it won't be confusing later

src/swagger/client.ts
KacperSB: thx!

KacperSB (approved):
LGTM

Kamil-Gumieniuk (approved):

KrzysztofNocen (approved):
LGTM

src/swagger/client/functional-testing-api.ts
bartoszkosicki13: To avoid another round of review i skip adding this comments, but in the next mcp task we can add it, if it is necessary.

PR #510, opened 2026-06-03
feat: add scan_api_standardization_from_registry tool

lukasikp:
## Goal

Reduce token usage when scanning APIs from the registry. Previously, the model had to fetch the full API definition (often a large spec) and pass it as raw text to the scan tool, which caused truncation.

## Design

A new `scan_api_standardization_from_registry` tool combines both steps: it fetches the definition directly from the registry using a `text/plain` accept header and runs the scan internally. The model only provides org/name/version and receives structured results, never handling the spec content itself.

## Changeset

- Added `scanApiStandardizationFromRegistry` MCP tool
- Added `text/plain` accept header support to `getApiDefinition`
- Structured scan response with `count`, `countsBySeverity`, and `url` fields

## Testing

Tested locally. Unit tests added for the new method.
image


kacperdufrat (approved):

szymon-pleszkun-sb (approved):

KrzysztofNocen (approved):

mzheng-sb:
this looks good, have we ensured that LLM will not be confused about which tool to use given that there are two scan tools which are very similar, is there any more wording change need to be done on description of the tool to make them distinctively clear about the two use cases?

lukasikp:
> this looks good, have we ensured that LLM will not be confused about which tool to use given that there are two scan tools which are very similar, is there any more wording change need to be done on description of the tool to make them distinctively clear about the two use cases?

Good point @mzheng-sb 
I tested this and the LLM does distinguish between the two tools based on the prompt context. If the prompt includes an API name, version, and org name, it picks the registry-based scan tool. If a spec is attached directly, it uses the other one. I also made slight improvements to the tool descriptions to make the two use cases clearer.

VarshaNagtilak1 (approved):

mzheng-sb (approved):

PR #520, opened 2026-06-12
feat(functional-testing): Run and query test execution

sergioifg94:
## Goal

Fixes [RF-4892](https://smartbear.atlassian.net/browse/RF-4892) and [RF-4925](https://smartbear.atlassian.net/browse/RF-4925)

Extend the Functional Testing toolset by adding support to running tests as well as querying the status and result of the execution.

Minor refactors, and support local development environment

## Design

Following the approach of `list_tests` as well as Reflect tool implementation

## Changeset

* Added methods in the Functional Testing API to run a test, and to get the execution status and results
* Added type definition for the input parameters
* Minor refactors:
  * Utility function to assert Functional Testing is configured
  * Support to use local development environment if `NODE_ENV` environment variable is set to `development`

## Testing

* Unit tests covering both features
* Ran locally and verified the functionality with Claude Desktop. See example:
  image

[RF-4892]: https://smartbear.atlassian.net/browse/RF-4892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RF-4925]: https://smartbear.atlassian.net/browse/RF-4925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server.json
+++ b/server.json
@@ -62,6 +62,12 @@
           "isSecret": true
         },
         {
+          "name": "SWAGGER_FUNCTIONAL_TESTING_BASE_PATH",
+          "description": "Override the Swagger Functional Testing API base URL (optional). Defaults to https://api.reflect.run/v1",
+          "isRequired": false,
+          "isSecret": false
+        },
+        {
           "name": "PACT_BROKER_BASE_URL",
           "description": "PactFlow/Pact Broker base URL. Leave empty to disable Pact tools. Learn more: https://developer.smartbear.com/smartbear-mcp/docs/contract-testing-with-pactflow",
           "isRequired": false,
--- a/src/swagger/client/functional-testing-api.ts
+++ b/src/swagger/client/functional-testing-api.ts
@@ -1,13 +1,23 @@
 import { ToolError } from "../../common/tools";
+import type {
+  GetFunctionalTestingExecutionTestParams,
+  RunFunctionalTestingTestParams,
+} from "./functional-testing-types";
 
 const API_HOSTNAME = "api.reflect.run";
+
 export const FUNCTIONAL_TESTING_API_KEY_HEADER = "X-API-KEY";
 
 export class FunctionalTestingAPI {
+  private baseUrl: string;
+
   constructor(
     private readonly getToken: () => string | null,
     private readonly userAgent: string,
-  ) {}
+    baseUrl?: string,
+  ) {
+    this.baseUrl = baseUrl || `https://${API_HOSTNAME}/v1`;
+  }
 
   getFtHeaders(): Record<string, string> {
     const token = this.getToken();
@@ -22,7 +32,7 @@ export class FunctionalTestingAPI {
   }
 
   async listTests(): Promise<unknown> {
-    const response = await fetch(`https://${API_HOSTNAME}/v1/tests`, {
+    const response = await fetch(`${this.baseUrl}/tests`, {
       method: "GET",
       headers: this.getFtHeaders(),
     });
@@ -30,6 +40,50 @@ export class FunctionalTestingAPI {
     if (!response.ok) {
       throw new ToolError(
         `Failed to list Functional Testing tests: ${response.status} ${response.statusText}`,
+      );
+    }
+
+    return response.json();
+  }
+
+  async runTest(args: RunFunctionalTestingTestParams): Promise<unknown> {
+    if (!args.testId) throw new ToolError("testId argument is required");
+
+    const response = await fetch(
+      `${this.baseUrl}/tests/${args.testId}/executions`,
+      {
+        method: "POST",
+        headers: this.getFtHeaders(),
+      },
+    );
+
+    if (!response.ok) {
+      throw new ToolError(
+        `Failed to run test: ${response.status} ${response.statusText}`,
+      );
+    }
+
+    return response.json();
+  }
+
+  async getTestExecution(
+    args: GetFunctionalTestingExecutionTestParams,
+  ): Promise<unknown> {
+    if (!args.executionId) {
+      throw new ToolError("executionId argument is required");
+    }
+
+    const response = await fetch(
+      `${this.baseUrl}/executions/${args.executionId}`,
+      {
+        method: "GET",
+        headers: this.getFtHeaders(),
+      },
+    );
+
+    if (!response.ok) {
+      throw new ToolError(
+        `Failed to get test status: ${response.status} ${response.statusText}`,
       );
     }
 
--- a/src/swagger/client/functional-testing-tools.ts
+++ b/src/swagger/client/functional-testing-tools.ts
@@ -1,4 +1,8 @@
 import { z } from "zod";
+import {
+  GetFunctionalTestingExecutionTestSchema,
+  RunFunctionalTestingTestParamsSchema,
+} from "./functional-testing-types";
 import type { SwaggerToolParams } from "./tools";
 import { READ_ONLY } from "./tools";
 
@@ -20,5 +24,28 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
     ...READ_ONLY,
     handler: "listFunctionalTestingTests",
     outputSchema: ListTestsOutputSchema,
+  },
+  {
+    title: "Run Test",
+    toolset: "Functional Testing",
+    summary:
+      "Runs a specific API test in your Swagger Functional Testing workspace. " +
+      "The execution is asynchronous — it returns an executionId, not the result directly. " +
+      "Use swagger_get_test_status with that executionId to track progress and retrieve the final result.",
+    inputSchema: RunFunctionalTestingTestParamsSchema,
+    handler: "runFunctionalTestingTest",
+    idempotent: false,
+    readOnly: false,
+  },
+  {
+    title: "Get Test Status",
+    toolset: "Functional Testing",
+    summary:
+      "Get the status of a Swagger Functional Testing test execution. " +
+      "It returns information about the execution such as its status (running, passed or failed), run time, " +
+      "as well as the break down of the status of each test step.",
+    inputSchema: GetFunctionalTestingExecutionTestSchema,
+    handler: "getFunctionalTestingExecution",
+    idempotent: false,
   },
 ];
--- a/src/swagger/client/functional-testing-types.ts
+++ b/src/swagger/client/functional-testing-types.ts
@@ -1,0 +1,24 @@
+import { z } from "zod";
+
+export const RunFunctionalTestingTestParamsSchema = z.object({
+  testId: z
+    .string()
+    .describe("ID of the Functional Testing test to run")
+    .trim()
+    .min(1),
+});
+
+export const GetFunctionalTestingExecutionTestSchema = z.object({
+  executionId: z
+    .string()
+    .describe("ID of the Functional Testing execution")
+    .trim()
+    .min(1),
+});
+
+export type RunFunctionalTestingTestParams = z.infer<
+  typeof RunFunctionalTestingTestParamsSchema
+>;
+export type GetFunctionalTestingExecutionTestParams = z.infer<
+  typeof GetFunctionalTestingExecutionTestSchema
+>;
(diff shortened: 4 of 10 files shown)


src/swagger/client/functional-testing-api.ts
```diff
@@ -35,4 +43,66 @@ export class FunctionalTestingAPI {
...
+   * Get the URL base for API calls. When running against a local environment (NODE_ENV: development),
+   * the endpoints are exposed at the `/api` path. When running against production, the endpoints
+   * are exposed at `/v1`.
+   *
+   * If NODE_ENV is anything other than `development`, defaults to the production URL
+   */
+  private getUrlBase(): string {
+    if (!this.urlBase) {
+      this.urlBase =
+        process.env.NODE_ENV === "development"
```
sergioifg94: Haven't found another existing mechanism to tell whether we're running locally (as in development environment) or not. If there isn't, it might be interesting to extend this at some point to have a unified approach

src/swagger/client/functional-testing-tools.ts
```diff
@@ -10,4 +14,20 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
       "Do not use this tool to retrieve test execution results or history.",
     handler: "listFunctionalTestingTests",
   },
+  {
+    title: "Run Test",
+    toolset: "Functional Testing",
+    summary:
+      "Runs a specific API test in your Swagger Functional Testing workspace and returns the result. " +
```
KacperSB: In `swagger-functional-testing-integration.mc` you wrote:

> Returns: Execution details including an `executionId` that can be used to poll for the result.

If thats the case, then the tool does not return test result, so this line needs adjustment

src/swagger/client/functional-testing-types.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from "zod";
+
+export const RunFunctionalTestingTestParamsSchema = z.object({
+  testId: z.string().describe("ID of the Functional Testing test to run"),
+});
+
+export const GetFunctionalTestingExecutionSchema = z.object({
+  executionId: z.string().describe("ID of the Functional Testing execution"),
+});
```
KacperSB, line 17: We could expand these `zod` schemas a little, at least a `.min(1)` on `testId` and on `executionId` since from what i see these are required args.

src/swagger/client/functional-testing-api.ts
```diff
@@ -35,4 +43,66 @@ export class FunctionalTestingAPI {
...
+          ? `http://${API_HOSTNAME_DEV}/api`
+          : `https://${API_HOSTNAME}/v1`;
+    }
```
KacperSB: Optionally you could move this into `constructor` at lines 15-18, so that env can be injectable for tests.

```ts
private readonly urlBase = process.env.NODE_ENV === "development"
      ? `http://${API_HOSTNAME_DEV}/api`
      : `https://${API_HOSTNAME}/v1`
```

src/swagger/client/functional-testing-tools.ts
```diff
@@ -10,4 +14,20 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
...
+      "Use this tool when you need to verify expected API functionality by executing a single test.",
+    inputSchema: RunFunctionalTestingTestParamsSchema,
+    handler: "runFunctionalTestingTest",
+  },
+  {
+    title: "Get Test Status",
+    toolset: "Functional Testing",
+    summary: "Get the status of a Swagger Functional Testing test execution",
```
KacperSB: I find this summary little poor, maybe it should contain possible statuses (copleted?, failed?, passed?)

KacperSB (commented):
Some non-blocking suggestions:

src/swagger/client/functional-testing-tools.ts
bartoszkosicki13: I would give some additional information about next potential step that llm can go:
```
"Runs a specific API test in your Swagger Functional Testing workspace. 
The execution is asynchronous — it returns an executionId, not the result directly. 
Use swagger_get_test_status with that executionId to track progress and retrieve the final result."
```

src/swagger/client/functional-testing-api.ts
bartoszkosicki13: I can see NODE_ENV is used here to point at a local stack during development- makes sense, but maybe we could use mechanism for exactly this in SwaggerConfiguration. All other APIs (Portal, Registry, UserManagement) use overridable base paths with a production default:

configuration.ts - existing pattern
this.portalBasePath = param.portalBasePath || "https://api.portal.swaggerhub.com/v1";

Would suggest adding ftBasePath to SwaggerConfigurationParameters and passing it into the FunctionalTestingAPI constructor- local dev then works the same way as for the other APIs (override via config), and NODE_ENV, API_HOSTNAME_DEV and the lazy urlBase cache can all be dropped.

I personally prefer environment to come from client config rather than a build-time variable. 
What do you think?

Kamil-Gumieniuk: Yes, agree to @bartoszkosicki13 suggestion. We can add an optional functional_testing_base_url to ConfigurationSchema, pass it into FunctionalTestingAPI, and default to https://api.reflect.run/v1 when unset. This way we can dropping the NODE_ENV branch entirely.

src/swagger/client/functional-testing-tools.ts
```diff
@@ -10,4 +14,20 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
...
     handler: "listFunctionalTestingTests",
   },
+  {
+    title: "Run Test",
+    toolset: "Functional Testing",
+    summary:
+      "Runs a specific API test in your Swagger Functional Testing workspace and returns the result. " +
+      "Use this tool when you need to verify expected API functionality by executing a single test.",
+    inputSchema: RunFunctionalTestingTestParamsSchema,
+    handler: "runFunctionalTestingTest",
```
Kamil-Gumieniuk, line 36: We should set `readOnly` and `idempotent` parameters here. Defaults to them is `true` and it's not valid for this `run_test`

Kamil-Gumieniuk (commented):
I agree with @bartoszkosicki13 regarding changes for mechanism for urlBase (details in separate comment). Besides that, I would simply advice to set readOnly / idempotent params. I have no other comments than already reported.

src/swagger/client/functional-testing-api.ts
sergioifg94: I had not considered that possibility, thank you for raising it @bartoszkosicki13 @Kamil-Gumieniuk, I made the changes now to set the base url in the configuration, following the pattern of other services. Let me know what you think

src/swagger/client/functional-testing-tools.ts
sergioifg94: Good catch @KacperSB that was out of date, and thanks @bartoszkosicki13, I followed your suggestion and used that as the summary

src/swagger/client/functional-testing-types.ts
sergioifg94: I might be wrong but my understanding is that the fields are required by default and, in order to make them optional, we'd need to call `.optional()`

src/swagger/client/functional-testing-api.ts
sergioifg94: Moved the url base to the constructor which is passed now from the configuration

src/swagger/client/functional-testing-tools.ts
sergioifg94: I expanded the summary with more information, let me know what you think

src/swagger/client/functional-testing-tools.ts
sergioifg94: Good call, I wasn't aware of those parameters, I set them in the last commit

src/swagger/client/functional-testing-api.ts
```diff
@@ -1,13 +1,23 @@
 import { ToolError } from "../../common/tools";
+import type {
+  GetFunctionalTestingExecutionParams,
```
KacperSB: Could we rename to:
```suggestion
  GetFunctionalTestingExecutionTestParams,
```
?
Soon we will have suites, so would be nice to distinguish them

src/swagger/client/functional-testing-api.ts
```diff
@@ -35,4 +45,48 @@ export class FunctionalTestingAPI {
...
+    if (!response.ok) {
+      throw new ToolError(
+        `Failed to run test: ${response.status} ${response.statusText}`,
+      );
+    }
+
+    return response.json();
+  }
+
+  async getExecution(
```
KacperSB: Same here, maybe we could do:

```suggestion
  async getTestExecution(
```
so later for suites we will have `getSuiteExecution`

src/swagger/client/functional-testing-tools.ts
```diff
@@ -10,4 +14,23 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
       "Do not use this tool to retrieve test execution results or history.",
     handler: "listFunctionalTestingTests",
   },
+  {
+    title: "Run Test",
+    toolset: "Functional Testing",
+    summary:
+      "Runs a specific API test in your Swagger Functional Testing workspace. The execution is asynchronous — it returns an executionId, not the result directly.  Use swagger_get_test_status with that executionId to track progress and retrieve the final result.",
```
KacperSB: [nit]
```suggestion
      "Runs a specific API test in your Swagger Functional Testing workspace. The execution is asynchronous — it returns an executionId, not the result directly. Use swagger_get_test_status with that executionId to track progress and retrieve the final result.",
```

src/swagger/client/functional-testing-tools.ts
```diff
@@ -10,4 +14,23 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
       "Do not use this tool to retrieve test execution results or history.",
     handler: "listFunctionalTestingTests",
   },
+  {
+    title: "Run Test",
```
KacperSB, line 29: [optional] To improve readability, for this tool and below one, you could split the summary in sentences, just like it's done above for `List Tests`

src/swagger/client/functional-testing-types.ts
KacperSB: @sergioifg94 I might explained that poorly, it's not about ommiting the string, i'ts rather when you pass it as empty "", and it doesn't matter if it's optional value or not. Without `min(1)` (minimal length of string = 1) LLM could sent "" and zod would not reject it. I know it sounds insane to pass "" for human, but for AI that's a guard worth to have

KacperSB (approved):
Some suggestions, good to go overall

server.json
```diff
@@ -61,6 +61,12 @@
           "isRequired": false,
           "isSecret": true
         },
+        {
+          "name": "SWAGGER_FUNCTIONAL_TESTING_BASE_PATH",
+          "description": "Override the Swagger Functional Testing API base URL (optionsl). Defaults to https://api.reflect.run/v1",
```
bartoszkosicki13: here is typo (optionsl) -> (optional)

bartoszkosicki13 (approved):
LGTM

KacperSB (commented):
@sergioifg94  Here's a review from @mzheng-sb that he added on my PR, but should be there:

Image

src/swagger/client/functional-testing-api.ts
sergioifg94: Good idea, I updated it

src/swagger/client/functional-testing-api.ts
sergioifg94: And this 👌

src/swagger/client/functional-testing-types.ts
sergioifg94: Nice, I got it now. I added `.trim().min(1)` to validate also for things like `"   "`

Kamil-Gumieniuk (approved):
All  previously mentioned things were fixed. Nothing more on my side. Nice work!

KrzysztofNocen (approved):

bartoszkosicki13 (approved):

PR #525, opened 2026-06-16
feat: update swagger_publish_portal_product tool to get the live and preview urls for page or product (SWG-20373)

VarshaNagtilak1:
## Goal
Ticket: SWG-20373
Fix swagger_publish_portal_product so the tool itself returns the live or preview page URL. This removes the need for the agent skill or LLM to construct URLs, which was causing incorrect links.

## Design

Generate the publish URL inside the MCP tool using the actual product, portal, section, and optional [tableOfContentsId] data returned from the API. Keep URL logic in the tool response so the agent can display the exact link directly without inference.

## Changeset

Updated swagger_publish_portal_product to accept [tableOfContentsId], resolve the matching TOC item when provided, and return the computed live or preview page URL in the tool response. Also narrowed the returned metadata to only the fields needed for the publish result.

## Testing

The unit tests for [publishPortalProduct] were updated to cover:

live publish with [tableOfContentsId]
preview publish without [tableOfContentsId]
custom domain URL generation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- [Swagger] Refactored `publish_portal_product` to build live URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` in the publish response.
+[#525](https://github.com/SmartBear/smartbear-mcp/pull/525)
+
 ## [0.26.1] - 2026-06-24
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Fixed
+### Added
 
-- [Swagger] Refactored `publish_portal_product` to build published URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` or `previewUrl` plus product, portal, and table-of-contents metadata in the publish response. If `portal.customDomain` is present, the client now uses it as the full host without appending the portal UI suffix.
+- [Swagger] Extended `publish_portal_product` to build published URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` or `previewUrl` plus product, portal, and table-of-contents metadata in the publish response. If `portal.customDomain` is present, the client now uses it as the full host without appending the portal UI suffix.
 [#525](https://github.com/SmartBear/smartbear-mcp/pull/525)
 
 ## [0.26.1] - 2026-06-24
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- [Swagger] Refactored `publish_portal_product` to build live URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` in the publish response.
+- [Swagger] Refactored `publish_portal_product` to build published URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` or `previewUrl` in the publish response.
 [#525](https://github.com/SmartBear/smartbear-mcp/pull/525)
 
 ## [0.26.1] - 2026-06-24
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- [Swagger] Refactored `publish_portal_product` to build published URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` or `previewUrl` in the publish response.
+- [Swagger] Refactored `publish_portal_product` to build published URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` or `previewUrl` plus product, portal, and table-of-contents metadata in the publish response. If `portal.customDomain` is present, the client now uses it as the full host without appending the portal UI suffix.
 [#525](https://github.com/SmartBear/smartbear-mcp/pull/525)
 
 ## [0.26.1] - 2026-06-24
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


docs/products/SmartBear MCP Server/swagger-portal-integration.md
```diff
@@ -116,14 +116,28 @@ The Swagger Portal client provides comprehensive portal and product management c
...
 
 | Parameter   | Description                                                                                                                          | Type    | Required |
 | ----------- | ------------------------------------------------------------------------------------------------------------------------------------ | ------- | -------- |
 | `productId` | Product UUID or identifier in the format 'portal-subdomain:product-slug' - unique identifier for the product                         | string  | Yes      |
 | `preview`   | Whether to publish as preview (true) or live (false). Preview allows testing before going live. Defaults to false (live publication) | boolean | No       |
+| `tableOfContentsId` | Optional table of contents UUID, or identifier in the format 'portal-subdomain:product-slug:section-slug:table-of-contents-slug' used to resolve a specific live URL path | string | No |
+
+**Response Details:**
+- When `portal.customDomain` is present, it is used as the full host in the generated URL without appending the portal UI suffix
+- When `customDomain` is not present, the URL is built using `portal.subdomain` + the portal UI suffix (`.portal.swaggerhub.com`)
```
lukasikp: Those first two bullet points explain how the URL is built internally (custom domain vs subdomain). The tool already returns the final URL, so the caller does not need to know this. This is an implementation detail, not user-facing docs. Safe to remove.

docs/products/SmartBear MCP Server/swagger-portal-integration.md
```diff
@@ -116,14 +116,28 @@ The Swagger Portal client provides comprehensive portal and product management c
...
+- When `tableOfContentsId` resolves successfully, the returned URL includes the section and table-of-contents path in the format: `/{productSlug}/{sectionSlug}/{tocSlug}`
```
lukasikp: This section describes implementation details (URL construction logic) rather than user-facing behavior. Replaced with a simpler description of what the response contains.
```suggestion
- The response contains a URL to the product. If tableOfContentsId is found, the URL links directly to that table of contents page.
```

src/swagger/client/tools.ts
```diff
@@ -107,7 +107,7 @@ export const TOOLS: SwaggerToolParams[] = [
     title: "Publish Portal Product",
     toolset: "Products",
     summary:
-      "Publish a product's content to make it live or as preview. This endpoint publishes the current content of a product, making it visible to portal visitors. Use preview mode to test before going live.",
+      "Publish a product's content to make it live or as preview. This endpoint publishes the current content of a product, making it visible to portal visitors. Use preview mode to test before going live. Optionally provide tableOfContentsId to form liveUrl or previewUrl for specific table-of-contents item or specific page url. Returns a PublishPortalProductResponse containing: success status, preview boolean, liveUrl or previewUrl (depending on preview mode), and narrowed metadata objects for product (id, name, slug), portal (id, name, subdomain, customDomain), and tableOfContentsItem (id, slug, title, order, parentId) when resolved.",
```
lukasikp: maybe something like this ? 
```suggestion
      "Publish a product's content to make it live or as preview. This endpoint publishes the current content of a product, making it visible to portal visitors. Use preview mode to test before going live. Optionally provide tableOfContentsId to get a live or preview URL pointing to a specific page. Returns publication status, a live or preview URL, and metadata for the product and portal.",
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- [Swagger] Refactored `publish_portal_product` to build published URLs dynamically from `SWAGGER_PORTAL_BASE_PATH`, support preview and section/table-of-contents paths, and return the resolved `liveUrl` or `previewUrl` plus product, portal, and table-of-contents metadata in the publish response. If `portal.customDomain` is present, the client now uses it as the full host without appending the portal UI suffix.
```
lukasikp: I think this point should be in the ### Added section, because we are extending the endpoint, not fixing it. I would also slightly modify the description: "Refactored" → "Extended".

lukasikp (approved):
good work. I just have a few minor suggestions for simplifying the descriptions.

mzheng-sb (approved):

src/swagger/client/api.ts
```diff
@@ -355,20 +357,129 @@ export class SwaggerAPI {
...
+   * @param tableOfContentsId - Optional table of contents UUID, or identifier in the format 'portal-subdomain:product-slug:section-slug:table-of-contents-slug'
+   * @returns Complete publish response with product details and the resolved published URL
+   */
   async publishPortalProduct(
     productId: string,
     preview: boolean = false,
-  ): Promise<SuccessResponse | FallbackResponse> {
+    tableOfContentsId: string | null = null,
+  ): Promise<PublishPortalProductResponse | FallbackResponse> {
+    const metadata = await this.preparePublicationMetadata(
```
KrzysztofNocen: **[High] Publish is now gated on the metadata GETs.**

`preparePublicationMetadata()` runs before the publish `PUT`, and `handleResponse` throws on any non-2xx response. So if `getPortalProduct`, `getPortalProductSections`, or `getPortal` fails (404, transient 5xx, missing data), the actual publish never happens and the whole tool errors out.

Previously the publish required only the `PUT`. This is a behavioral regression: the mutating action the user actually requested is now blocked by URL-building, which is cosmetic by comparison.

Suggestion: either publish first and build the URL afterward, or wrap metadata resolution in try/catch and return the successful publish with an omitted/empty URL on failure. URL generation should be best-effort, not a precondition for publishing.

src/swagger/client/api.ts
```diff
@@ -355,20 +357,129 @@ export class SwaggerAPI {
...
+      this.getPortalProductSections(productId, {
+        embed: ["tableOfContents", "tableOfContents.swaggerhubApi"],
+      }),
+    ]);
+
+    const portalDetails = await this.getPortal(
+      String(productDetails.portalId ?? ""),
+    );
+
+    const targetSection = sections.items[0] ?? null;
```
KrzysztofNocen: **[Medium] TOC lookup only searches the first section.**

`findTableOfContentsItem` is only run against `items[0]`. If a product has multiple sections and the `tableOfContentsId` belongs to any section other than the first, it silently won't resolve - `tableOfContentsItem` is omitted and the URL falls back to the product root with no indication the requested page wasn't found.

Consider iterating sections to find the one that actually contains the TOC item, and use that section's slug for the path.

src/swagger/client/api.ts
```diff
@@ -355,20 +357,129 @@ export class SwaggerAPI {
...
+  }> {
+    const [productDetails, sections] = await Promise.all([
+      this.getPortalProduct(productId),
+      this.getPortalProductSections(productId, {
+        embed: ["tableOfContents", "tableOfContents.swaggerhubApi"],
+      }),
+    ]);
+
+    const portalDetails = await this.getPortal(
+      String(productDetails.portalId ?? ""),
```
KrzysztofNocen: **[Medium] `getPortal("")` when `portalId` is missing.**

If `productDetails.portalId` is absent, `String(productDetails.portalId ?? "")` requests `/portals/` (empty id), which will fail - and per the High finding, that failure takes the whole publish down. Guard against an empty `portalId` explicitly before calling `getPortal`.

KrzysztofNocen (commented):
Two findings from review - one High (publish ordering/error handling) and two Medium (TOC multi-section lookup, empty portalId). Details inline.

src/swagger/client/api.ts
VarshaNagtilak1: This is really good catch!!! Thanks for this :)

src/swagger/client/api.ts
VarshaNagtilak1: In portal they are always using first section from product section. So its aligned with portal ui. In BE there can be multiple sections but Portal FE always uses items[0]

src/swagger/client/api.ts
VarshaNagtilak1: Done.

KrzysztofNocen (approved):

lukasikp (approved):

lukasikp (approved):

PR #541, opened 2026-06-24
feat(functional-testing): List suites

rafal-pawlow:
## Goal

Implements [RF-4890](https://smartbear.atlassian.net/browse/RF-4890).  As a Developer or QA Engineer, I want to list all test suites in my Swagger Functional Testing workspace via MCP, so that I can discover available suites before running or managing them.

## Design

Followed the existing `list_tests` pattern in the Swagger Functional Testing client. Added a `listSuites()` method on `FunctionalTestingAPI` that calls `GET https://api.reflect.run/v1/suites` with the `X-API-KEY` header, a `listFunctionalTestingSuites()` delegate on `SwaggerClient`, and a new `List Suites` tool entry gated behind `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` (same conditional registration as `list_tests`).

Per the ticket's negative-path requirements, failures map to actionable messages instead of the generic error used by `list_tests`:
- `401`/`403` → "Authentication failed. Verify your API token is valid and has not expired."
- Other non-OK responses and network failures → "Swagger Functional Testing service is currently unreachable. Retry after a moment."

An empty workspace returns `[]` and is treated as a successful result, not an error.

## Changeset

- `src/swagger/client/functional-testing-api.ts` — added `listSuites()` with scoped error handling.
- `src/swagger/client.ts` — added `listFunctionalTestingSuites()` delegate.
- `src/swagger/client/functional-testing-tools.ts` — registered the `List Suites` tool.
- `docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md` — documented `list_suites`.
- `CHANGELOG.md` — added entry under `Unreleased`.
- `src/tests/unit/swagger/functional-testing/*` — added API- and client-level tests.

## Testing

- [X] Existing unit tests pass
- [X] Extended unit tests: `functional-testing-api.test.ts` and `swagger-client-ft.test.ts`

[RF-4890]: https://smartbear.atlassian.net/browse/RF-4890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Swagger] Register Portal/Studio tools when authenticated via OAuth. Previously only an explicit API key (`SWAGGER_API_KEY` / `Swagger-Api-Key` header) enabled the tools, so OAuth-only sessions listed no Swagger tools. [#537](https://github.com/SmartBear/smartbear-mcp/pull/537)
 
+### Added
+
+- [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
+
 ## [0.26.0] - 2026-06-19
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Pactflow]: Send the MCP client's name and version to Pactflow via a `SOURCE_APPLICATION` header on requests, captured from the client info supplied in the MCP `initialize` request. [#556](https://github.com/SmartBear/smartbear-mcp/pull/556)
 
+### Added
+
+- [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
+
 ### Fixed
 
 - [Zephyr] Fixed `get_test_case_steps` returning a schema validation error [#543](https://github.com/SmartBear/smartbear-mcp/pull/543)
@@ -26,8 +30,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Swagger] Add `create_documentation_page` tool to create a documentation page in a portal product in a single call. Supports `markdown` and `html` content types with `internal` or `external` source. Returns page details and a `draftUrl` to edit the page in the portal admin.
 
 - [Swagger] Extended Swagger Functional Testing integration with `run_test`, and `get_test_status` tools for executing available API tests and querying their execution.
-
-- [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
 
 ## [0.26.1] - 2026-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,15 +27,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Swagger] Extended Swagger Functional Testing integration with `run_test`, and `get_test_status` tools for executing available API tests and querying their execution.
 
+- [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
+
 ## [0.26.1] - 2026-06-24
 
 ### Fixed
 
 - [Swagger] Register Portal/Studio tools when authenticated via OAuth. Previously only an explicit API key (`SWAGGER_API_KEY` / `Swagger-Api-Key` header) enabled the tools, so OAuth-only sessions listed no Swagger tools. [#537](https://github.com/SmartBear/smartbear-mcp/pull/537)
-
-### Added
-
-- [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
 
 ## [0.26.0] - 2026-06-19
 
--- a/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
+++ b/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
@@ -15,7 +15,7 @@ The Swagger Functional Testing client provides tools for discovering and executi
 #### `list_suites`
 
 - Purpose: Lists all test Suites available in your Swagger Functional Testing workspace. Use this tool when you need to discover available Suites before running them or checking their execution history. Do not use this tool to retrieve individual tests or test Suite execution results.
-- Returns: Complete list of test Suites in the workspace. An empty list (`[]`) is returned when no Suites exist.
+- Returns: An object with a `suites` array of the test Suites in the workspace, alongside aggregate `stats`. When no Suites exist, the `suites` array is empty (`{ "suites": [] }`).
 - Use case: Discover available test Suites.
 
 ---
--- a/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
+++ b/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
@@ -12,6 +12,12 @@ The Swagger Functional Testing client provides tools for discovering and executi
 - Returns: Complete list of tests with their identifiers and names.
 - Use case: Discover available tests.
 
+#### `list_suites`
+
+- Purpose: Lists all test suites available in your Swagger Functional Testing workspace. Use this tool when you need to discover available suites before running them or checking their execution history. Do not use this tool to retrieve individual tests or test suite execution results.
+- Returns: Complete list of test suites in the workspace. An empty list (`[]`) is returned when no suites exist.
+- Use case: Discover available test suites.
+
 ---
 
 ### Test Execution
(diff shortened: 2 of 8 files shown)


KacperSB:
I will review later once https://github.com/SmartBear/smartbear-mcp/pull/542 is merged and this PR is rebased

rafal-pawlow:
@KacperSB Done

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
```diff
@@ -12,6 +12,12 @@ The Swagger Functional Testing client provides tools for discovering and executi
 - Returns: Complete list of tests with their identifiers and names.
 - Use case: Discover available tests.
 
+#### `list_suites`
+
+- Purpose: Lists all test suites available in your Swagger Functional Testing workspace. Use this tool when you need to discover available suites before running them or checking their execution history. Do not use this tool to retrieve individual tests or test suite execution results.
```
KacperSB: [nit] I think it's more readable when we capitalize (S)uites

src/swagger/client/functional-testing-api.ts
```diff
@@ -89,4 +90,32 @@ export class FunctionalTestingAPI {
...
+    if (response.status === 401 || response.status === 403) {
+      throw new ToolError(
+        "Authentication failed. Verify your API token is valid and has not expired.",
+      );
+    }
+
+    if (!response.ok) {
+      throw new ToolError(
+        "Swagger Functional Testing service is currently unreachable. Retry after a moment.",
+      );
```
KacperSB, line 117: Hmm, so we return same error when there is a problem with reaching suites URL and when the response is not OK. 
I would rather keep similar error like we have for all other tools for that case - which returns the response status and text to the client (i would rather want to know the response if it's not OK then see that the service is unavailable)

src/swagger/client/functional-testing-tools.ts
```diff
@@ -48,4 +48,13 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
...
     idempotent: false,
   },
+  {
+    title: "List Suites",
+    toolset: "Functional Testing",
+    summary:
+      "Lists all test suites available in your Swagger Functional Testing workspace. " +
+      "Use this tool when you need to discover available suites before running them or checking their execution history. " +
+      "Do not use this tool to retrieve individual tests or test suite execution results.",
+    handler: "listFunctionalTestingSuites",
```
KacperSB, line 47: [optional] you could specify here params 'readOnly' and 'idempotent'. I know defaults for these (true) work for this tool, but IMO it's better to define them as the defaults can change any time in future

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Swagger] Register Portal/Studio tools when authenticated via OAuth. Previously only an explicit API key (`SWAGGER_API_KEY` / `Swagger-Api-Key` header) enabled the tools, so OAuth-only sessions listed no Swagger tools. [#537](https://github.com/SmartBear/smartbear-mcp/pull/537)
 
+### Added
+
+- [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
+
```
KacperSB: This should be under 'Added' but in 'Unreleased' section

src/swagger/client.ts
```diff
@@ -378,6 +378,13 @@ export class SwaggerClient implements Client {
     return fn(this.ftApi);
   }
 
+  async listFunctionalTestingSuites(): Promise<unknown> {
+    if (!this.ftApi) {
+      throw new ToolError("Functional Testing API not configured");
+    }
```
KacperSB: We have an existing handler for this logic: `withFunctionalTesting` (`swagger/client.ts:371`) that you should use to be consistent

src/swagger/client/functional-testing-types.ts
```diff
@@ -22,3 +22,22 @@ export type RunFunctionalTestingTestParams = z.infer<
...
+}
+
+export interface ListSuitesResponse {
+  suites: Suite[];
+  stats: {
+    executions: number;
+    passRate: number;
+    avgRuntimeSecs: number;
+    cumExecTimeSecs: number;
+  };
```
KacperSB, line 42: hmm, i think it would be better if you would define a `zod` output schema in a same way like we already did for rest of the tools.

src/tests/unit/swagger/functional-testing/functional-testing-api.test.ts
```diff
@@ -10,6 +10,11 @@ const testsMock = [
   { id: "test-2", name: "Checkout Test" },
 ];
 
+const suitesMock = [
+  { id: "suite-1", name: "Smoke Suite" },
+  { id: "suite-2", name: "Regression Suite" },
```
KacperSB: Suite `id` is defined as `number` in new interface you added but here it's mocked as string..
I think it should be defined as `string`, as we allow chars to be present in suite ID

src/swagger/client/functional-testing-types.ts
```diff
@@ -22,3 +22,22 @@ export type RunFunctionalTestingTestParams = z.infer<
...
+  numTestInstances: number;
+}
+
+export interface ListSuitesResponse {
+  suites: Suite[];
+  stats: {
+    executions: number;
+    passRate: number;
+    avgRuntimeSecs: number;
+    cumExecTimeSecs: number;
```
KacperSB: When user does not have any suites, the `suites` will be empty array, but what should happen to `stats`?Shouldn't these `stats` be optional to even allow response to be correct in such case?

KacperSB (changes requested):
Few things, good to go from me if those will be updated or addressed
Also please add Krzysztof or Paulina for review as they know the repo well

src/swagger/client/functional-testing-types.ts
rafal-pawlow: We already use ts interfaces in https://github.com/SmartBear/smartbear-mcp/blob/main/src/swagger/client/user-management-types.ts for the internal API shape and zod schemas specifically for the MCP outputSchema. I think my approach is valid and consistent.

KacperSB: I don't have an issue with these interfaces, just wondering if schema isn't needed here on top of them, if not that's fine

src/tests/unit/swagger/functional-testing/functional-testing-api.test.ts
```diff
@@ -10,6 +10,11 @@ const testsMock = [
...
+];
+
```
KacperSB, line 39: `functional-testing-types.ts:35-43` defines suites response as:
```ts  
export interface ListSuitesResponse {
    suites: Suite[];
    stats?: { ... };
  }
```

But in test we mock only the `suites` part, you could change it to verify full response shape, with `stats` included.

KacperSB: In that scope i think the mock could be also expanded to in fact verify all the fields of `suites`, not only `id` and `name`.

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Swagger] Extended Swagger Functional Testing integration with `run_test`, and `get_test_status` tools for executing available API tests and querying their execution.
 
+- [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
+
```
KacperSB: I still think this should be under `Unreleased`, since we did not yet released the `list_suites` tool. If i am wrong there please explain

KacperSB (approved):
Nice!
I left some minors, test improvements and (i believe) still wrong placement in changelog file

Asmodeios (approved):

src/swagger/client/functional-testing-api.ts
```diff
@@ -89,4 +90,32 @@ export class FunctionalTestingAPI {
 
     return response.json();
   }
+
+  async listSuites(): Promise<ListSuitesResponse> {
+    let response: Response;
+    try {
+      response = await fetch(`${this.baseUrl}/suites`, {
+        method: "GET",
+        headers: this.getFtHeaders(),
```
Kamil-Gumieniuk: `this.getFtHeaders()` can throw `ToolError("Swagger Functional Testing API token not found")` and here it is swallowed and replaced with generic
```
ToolError(
        "Swagger Functional Testing service is currently unreachable. Retry after a moment.",
      );
```
Can we read
```
const headers = this.getFtHeaders();
```
before try - catch instead?

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
```diff
@@ -12,6 +12,12 @@ The Swagger Functional Testing client provides tools for discovering and executi
 - Returns: Complete list of tests with their identifiers and names.
 - Use case: Discover available tests.
 
+#### `list_suites`
+
+- Purpose: Lists all test Suites available in your Swagger Functional Testing workspace. Use this tool when you need to discover available Suites before running them or checking their execution history. Do not use this tool to retrieve individual tests or test Suite execution results.
+- Returns: Complete list of test Suites in the workspace. An empty list (`[]`) is returned when no Suites exist.
```
Kamil-Gumieniuk: Descrioption regarding returning empty list is confusing. 
According to 
`it("should return an empty suites list when no suites exist", async () => {`
it returns object that contains empty list
`{ suites: [] }`

Kamil-Gumieniuk (approved):
Just minor comments from my side.

src/swagger/client/functional-testing-api.ts
rafal-pawlow: Good catch 👍

KacperSB (approved):
Just fyi i confirmed that this tool does not return any URLs we wouldn't want = )

PR #543, opened 2026-06-26
TM4J-17703: Zephyr get test case steps fails with schema validation error

BenzoEnzo:
## Goal

Regenerate the Zephyr REST API schemas to fix the `get_test_case_steps` tool failing with a schema validation error.

## Changeset

  - Regenerate `rest-api-schemas.ts`.

## Testing
- Verify if updated tool is properly recognized and loaded by Copilot Agent, then tested it with different prompts.


src/tests/unit/zephyr/tool/test-cycle/update-test-cycle.test.ts
```diff
@@ -406,15 +406,21 @@ describe("UpdateTestCycle", () => {
         project: {
           id: 10006,
         },
-        jiraProjectVersion: 10006,
```
juro994, line 409: Sorry, maybe a stupid question, but I can see that in the past this was actually an object (like in your current PR), but it was changed to a number [here](https://github.com/SmartBear/smartbear-mcp/pull/442/changes), and now you're changing it to object again.

Do you know why in the PR that I've linked above, the schema was changed so `jiraProjectVersion` is a number? My point is, are you sure that reverting it back to be an `object` with an `id` property is a safe operation?

juro994 (approved):

PR #551, opened 2026-06-29
feat(functional-testing): list suite executions

Asmodeios:
## Goal
Implements RF-4891 (https://smartbear.atlassian.net/browse/RF-4891)

Extends the SFT toolset with `list_suite_executions`.
- `list_suite_executions` lists all executions for a given test suite by `suiteId`, for reviewing execution history and timings. Returns each execution's `executionId`, `status`, and `isFinished`.

## Design
New tool delegates through a `SwaggerClient` method, the API call lives in `FunctionalTestingAPI`, and the input schema is defined with Zod in the dedicated types file. Return shape is typed via dedicated `SuiteExecution` / `ListSuiteExecutionsResponse` interfaces. Tool is marked `readOnly`.

Maps Reflect API `404` to a clear suite-not-found message; falls back to a generic message for other HTTP errors. Empty execution lists are returned as-is.

## Changeset
- Added `list_suite_executions` tool to `FunctionalTestingAPI` and wired it through `SwaggerClient`
- Added `ListFunctionalTestingSuiteExecutionsSchema` Zod input schema (`suiteId`, trimmed, min 1)
- Added `SuiteExecution` and `ListSuiteExecutionsResponse` return-type interfaces; typed `listSuiteExecutions` return
- Added suite-not-found (404) handling with generic error fallback
- Updated docs (swagger-functional-testing-integration.md)
- Updated CHANGELOG.md
- Added `ftFetch` wrapper for all the function in `functional-testing-api.ts`

## Testing
- [x] Unit tests added
- [x] Manual local testing of new SFT tool performed
- [x] Manual local regression testing of other SFT tools performed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
+++ b/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
@@ -31,7 +31,7 @@ The Swagger Functional Testing client provides tools for discovering and executi
 #### `list_suite_executions`
 
 - Purpose: Lists all executions for a given test suite in your Swagger Functional Testing workspace. Use this tool when you need to review execution history and timings for a specific suite. Do not use this tool to retrieve the status of a single execution or individual test results. Requires a `suiteId`.
-- Returns: Complete list of executions for the given suite, in the order returned by the API. An empty list is returned when no executions exist.
+- Returns: Complete list of executions for the given suite. An empty list is returned when no executions exist.
 - Use case: Review the execution history and timings of a test suite.
 
 ---
--- a/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
+++ b/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
@@ -28,6 +28,12 @@ The Swagger Functional Testing client provides tools for discovering and executi
 - Returns: Execution status and result details for the given execution.
 - Use case: Poll for the outcome of a test run after calling `run_test`.
 
+#### `list_suite_executions`
+
+- Purpose: Lists all executions for a given test suite in your Swagger Functional Testing workspace. Use this tool when you need to review execution history and timings for a specific suite. Do not use this tool to retrieve the status of a single execution or individual test results. Requires a `suiteId`.
+- Returns: Complete list of executions for the given suite, in the order returned by the API. An empty list is returned when no executions exist.
+- Use case: Review the execution history and timings of a test suite.
+
 ---
 
 ## Additional Notes
--- a/src/swagger/client/functional-testing-api.ts
+++ b/src/swagger/client/functional-testing-api.ts
@@ -143,24 +143,10 @@ export class FunctionalTestingAPI {
     };
   }
   async listSuites(): Promise<ListSuitesResponse> {
-    const headers = this.getFtHeaders();
-    let response: Response;
-    try {
-      response = await fetch(`${this.baseUrl}/suites`, {
-        method: "GET",
-        headers,
-      });
-    } catch {
-      throw new ToolError(
-        "Swagger Functional Testing service is currently unreachable. Retry after a moment.",
-      );
-    }
-
-    if (response.status === 401 || response.status === 403) {
-      throw new ToolError(
-        "Authentication failed. Verify your API token is valid and has not expired.",
-      );
-    }
+    const response = await this.ftFetch(`${this.baseUrl}/suites`, {
+      method: "GET",
+      headers: this.getFtHeaders(),
+    });
 
     if (!response.ok) {
       throw new ToolError(
--- a/src/swagger/client/functional-testing-api.ts
+++ b/src/swagger/client/functional-testing-api.ts
@@ -134,11 +134,13 @@ export class FunctionalTestingAPI {
     return {
       ...data,
       executions: {
-        data: data.executions.data.map(({ executionId, status, isFinished }) => ({
-          executionId,
-          status,
-          isFinished,
-        })),
+        data: data.executions.data.map(
+          ({ executionId, status, isFinished }) => ({
+            executionId,
+            status,
+            isFinished,
+          }),
+        ),
       },
     };
   }
@@ -170,4 +172,3 @@ function suiteExecutionsErrorMessage(response: Response): string {
       return `Failed to list suite executions: ${response.status} ${response.statusText}`;
   }
 }
-
--- a/src/swagger/client/functional-testing-api.ts
+++ b/src/swagger/client/functional-testing-api.ts
@@ -71,7 +71,7 @@ export class FunctionalTestingAPI {
     if (!args.testId) throw new ToolError("testId argument is required");
 
     const response = await this.ftFetch(
-      `${this.baseUrl}/tests/${args.testId}/executions`,
+      `${this.baseUrl}/tests/${encodeURIComponent(args.testId)}/executions`,
       {
         method: "POST",
         headers: this.getFtHeaders(),
@@ -95,7 +95,7 @@ export class FunctionalTestingAPI {
     }
 
     const response = await this.ftFetch(
-      `${this.baseUrl}/executions/${args.executionId}`,
+      `${this.baseUrl}/executions/${encodeURIComponent(args.executionId)}`,
       {
         method: "GET",
         headers: this.getFtHeaders(),
@@ -117,7 +117,7 @@ export class FunctionalTestingAPI {
     if (!args.suiteId) throw new ToolError("suiteId argument is required");
 
     const response = await this.ftFetch(
-      `${this.baseUrl}/suites/${args.suiteId}/executions`,
+      `${this.baseUrl}/suites/${encodeURIComponent(args.suiteId)}/executions`,
       {
         method: "GET",
         headers: this.getFtHeaders(),
--- a/src/swagger/client/functional-testing-api.ts
+++ b/src/swagger/client/functional-testing-api.ts
@@ -33,8 +33,30 @@ export class FunctionalTestingAPI {
     };
   }
 
+  private async ftFetch(
+    input: string,
+    init: RequestInit,
+  ): Promise<Response> {
+    let response: Response;
+    try {
+      response = await fetch(input, init);
+    } catch {
+      throw new ToolError(
+        "Swagger Functional Testing service is currently unreachable. Retry after a moment.",
+      );
+    }
+
+    if (response.status === 401 || response.status === 403) {
+      throw new ToolError(
+        "Authentication failed. Verify your API token is valid and has not expired.",
+      );
+    }
+
+    return response;
+  }
+
   async listTests(): Promise<unknown> {
-    const response = await fetch(`${this.baseUrl}/tests`, {
+    const response = await this.ftFetch(`${this.baseUrl}/tests`, {
       method: "GET",
       headers: this.getFtHeaders(),
     });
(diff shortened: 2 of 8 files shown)


docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
```diff
@@ -28,6 +28,12 @@ The Swagger Functional Testing client provides tools for discovering and executi
 - Returns: Execution status and result details for the given execution.
 - Use case: Poll for the outcome of a test run after calling `run_test`.
 
+#### `list_suite_executions`
+
+- Purpose: Lists all executions for a given test suite in your Swagger Functional Testing workspace. Use this tool when you need to review execution history and timings for a specific suite. Do not use this tool to retrieve the status of a single execution or individual test results. Requires a `suiteId`.
+- Returns: Complete list of executions for the given suite, in the order returned by the API. An empty list is returned when no executions exist.
```
KacperSB: [optional] "in the order returned by the API" - Someone who will use it will not know what's the API order, maybe we can just type that the order will be from newest to oldest runs?

src/swagger/client/functional-testing-api.ts
```diff
@@ -89,4 +91,37 @@ export class FunctionalTestingAPI {
...
+  ): Promise<ListSuiteExecutionsResponse> {
+    if (!args.suiteId) throw new ToolError("suiteId argument is required");
+
+    const response = await fetch(
+      `${this.baseUrl}/suites/${args.suiteId}/executions`,
+      {
+        method: "GET",
+        headers: this.getFtHeaders(),
+      },
+    );
```
KacperSB: Maybe we can wrap that call with `try` and re-throw any exception as `ToolError`

src/swagger/client/functional-testing-api.ts
```diff
@@ -89,4 +91,37 @@ export class FunctionalTestingAPI {
...
+
+    if (!response.ok) {
+      throw new ToolError(suiteExecutionsErrorMessage(response));
+    }
+
+    return response.json();
+  }
+}
+
+function suiteExecutionsErrorMessage(response: Response): string {
```
KacperSB, line 278: I like this approach a lot

src/swagger/client/functional-testing-tools.ts
```diff
@@ -37,4 +38,15 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
...
+  {
+    title: "List Suite Executions",
+    toolset: "Functional Testing",
+    summary:
+      "Lists all executions for a given test suite in your Swagger Functional Testing workspace. " +
+      "Use this tool when you need to review execution history and timings for a specific suite. " +
+      "Do not use this tool to retrieve the status of a single execution or individual test results.",
+    inputSchema: ListFunctionalTestingSuiteExecutionsSchema,
+    handler: "listFunctionalTestingSuiteExecutions",
+    readOnly: true,
```
KacperSB, line 55: [optional] you could also add `idempotent: true`

src/swagger/client/functional-testing-types.ts
```diff
@@ -16,9 +16,34 @@ export const GetFunctionalTestingExecutionTestSchema = z.object({
...
 export type GetFunctionalTestingExecutionTestParams = z.infer<
   typeof GetFunctionalTestingExecutionTestSchema
 >;
+export type ListFunctionalTestingSuiteExecutionsParams = z.infer<
+  typeof ListFunctionalTestingSuiteExecutionsSchema
+>;
+
+export interface SuiteExecution {
+  executionId: number;
+  url: string;
```
KacperSB: This `url` will not work properly when we release Private Workspaces as it doesn't contain `accountId` to switch to if you are on different workspace. We have a ticket that tracks fixing that here https://smartbear.atlassian.net/browse/RF-5271 but it won't be done before we release Private Workspaces

Could you for now exclude it and add a comment that we will bring it back corrected in scope of RF-5271?

src/swagger/client/functional-testing-api.ts
```diff
@@ -89,4 +91,37 @@ export class FunctionalTestingAPI {
...
+  switch (response.status) {
+    // Defensive: the Reflect API currently returns 200 with an empty
+    // `executions.data` list for an unknown suiteId rather than a 404, so this
```
KacperSB, line 281: Wow, i believe this is a bug.
Could you create a bug ticket for us to track fixing this?

KacperSB (changes requested):
I'm only requesting changes as we cannot expose suite urls (details in comment)
Other things are minor

I'm also wondering about pagination here, since on our UI we paginate Suite executions. Did you check how it will work for this tool in case of multi-page? Will it return all executions or only the first page? I would expect it to return all of them, at least for now (we can discuss that with the team if you want)

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
Asmodeios: @KacperSB The thing is that Reflect API doesn't mention it sorts out by newest when returning suite executions and there is no field to sort by ourselves. That's why it's written "in the order returned by the API" instead of "newest"(as requested in the ticket). I could just remove the line mentioning order.

src/swagger/client/functional-testing-api.ts
Asmodeios: https://smartbear.atlassian.net/browse/RF-5325

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
KacperSB: But you can just call it for few suites and see, i believe it does sort from newest to oldest
Anyway its optional, you can do whatever you want here

Asmodeios: Also all executions are returned, without pages

src/swagger/client/functional-testing-api.ts
Asmodeios: @KacperSB Added common fetch wrapper `ftFetch` for `class FunctionalTestingAPI`, so we don't need to add same stuff to every endpoint call

src/swagger/client/functional-testing-api.ts
```diff
@@ -89,4 +110,37 @@ export class FunctionalTestingAPI {
...
     return response.json();
   }
+
+  async listSuiteExecutions(
+    args: ListFunctionalTestingSuiteExecutionsParams,
+  ): Promise<ListSuiteExecutionsResponse> {
+    if (!args.suiteId) throw new ToolError("suiteId argument is required");
+
+    const response = await this.ftFetch(
+      `${this.baseUrl}/suites/${args.suiteId}/executions`,
```
KacperSB: I have found that there is method `encodeURIComponent` that swagger uses
You could update this part to use is as well: `${this.baseUrl}/suites/${encodeURIComponent(args.suiteId)}/executions`

Might be slight scope expansion but other links from sft could use it as well:
  `${this.baseUrl}/tests/${encodeURIComponent(args.testId)}/executions`
  `${this.baseUrl}/executions/${encodeURIComponent(args.executionId)}`

KacperSB (approved):
nothing more, below is really optional

src/swagger/client/functional-testing-types.ts
```diff
@@ -16,9 +16,35 @@ export const GetFunctionalTestingExecutionTestSchema = z.object({
...
   typeof GetFunctionalTestingExecutionTestSchema
 >;
+export type ListFunctionalTestingSuiteExecutionsParams = z.infer<
+  typeof ListFunctionalTestingSuiteExecutionsSchema
+>;
+
+export interface SuiteExecution {
+  executionId: number;
+  // Will be brought back after https://smartbear.atlassian.net/browse/RF-5271 is done
+  // url: string;
```
Kamil-Gumieniuk, line 75: PR description says that url is also returned:
> list_suite_executions lists all executions for a given test suite by suiteId, for reviewing execution history and timings. Returns each execution's executionId, url, status, and isFinished.

Kamil-Gumieniuk: Btw. Is there a reason for hidding it in the interface? 
The response next is returned as:
```
return response.json()
```
There is no output schema or response format. TypeScript interface would have no impact on runtime. Or am I missing sth here?

Kamil-Gumieniuk (approved):
Looks good. Please just let me know about commented `url` field

src/swagger/client/functional-testing-types.ts
KacperSB: I always feel like Kamil has disabled other review comments = )

I requested `url` removal [[here]](https://github.com/SmartBear/smartbear-mcp/pull/551#discussion_r3493609600). From talk with Wojtek we decided to not show it in MCP as the urls for suites/tests doesn't have `accountId` switcher, which is needed for Private Workspaces.

We can of course discuss it if you disagree

src/swagger/client/functional-testing-types.ts
```diff
@@ -16,9 +16,35 @@ export const GetFunctionalTestingExecutionTestSchema = z.object({
...
 export type GetFunctionalTestingExecutionTestParams = z.infer<
   typeof GetFunctionalTestingExecutionTestSchema
 >;
+export type ListFunctionalTestingSuiteExecutionsParams = z.infer<
+  typeof ListFunctionalTestingSuiteExecutionsSchema
+>;
+
+export interface SuiteExecution {
+  executionId: number;
+  // Will be brought back after https://smartbear.atlassian.net/browse/RF-5271 is done
```
KacperSB, line 74: I just verified if that `url` is really removed and sadly it's not:

Image

From what i checked it's not enough o remove the `url` from that interface as in `src/reflect/tool/suites/list-suite-executions.ts:40` you return the full response anyway. Using:
```ts
const data = await response.json();
    return {
      content: [{ type: "text", text: JSON.stringify(data) }],
    };
```
Consider filtering it instead:\
```ts
const data = await response.json();
const filtered = {
    ...data,
    executions: {
      data: data.executions.data.map(
        ({
          executionId,
          status,
          isFinished,
        }: {
          executionId: number;
          status: string;
          isFinished: boolean;
        }) => ({ executionId, status, isFinished }),
      ),
    },
  };
  return {
    content: [{ type: "text", text: JSON.stringify(filtered) }],
  };
```

I know it's ugly but once we fix URLs we will remove that code anyway. And when you will be fixing that, please add comment for that code part "// Will be adjusted after https://smartbear.atlassian.net/browse/RF-5271 is done" or something similar as it will help later find places that need adjustments.

KacperSB (changes requested):

src/swagger/client/functional-testing-types.ts
Asmodeios: Yep, we have to filter out response.

KacperSB (approved):
Nice, i think you can move the ticket to 'Waiting for testing' and remove yourself from assignee, someone will check it out

(Some time ago we agreed to test the MCP Tickets on PR branches before merging to main - not sure if you were on PTO that day so just explaining that here. Reason for that is no stage + we are not entirely responsible for swagger mcp releases)

src/swagger/client/functional-testing-api.ts
```diff
@@ -91,26 +112,44 @@ export class FunctionalTestingAPI {
     return response.json();
   }
 
-  async listSuites(): Promise<ListSuitesResponse> {
```
bartoszkosicki13, line 106: Why suites listing was removed?

Asmodeios: It's not removed, just moved when solving merge conflicts https://github.com/SmartBear/smartbear-mcp/pull/551/changes#diff-875885535559cc793ebacf3c3ac4b8cd830926b6b1dcb5fec00b4551f7bb6518R147

bartoszkosicki13: Sorry, I didn't update my local branch 👍 All good to me

bartoszkosicki13 (approved):

KacperSB (approved):
:shipit:

PR #553, opened 2026-06-29
feat: add output schema

bartosz-nowacki:
## Goal

ChatGPT Apps flagged all 26 Swagger MCP tools with: "Recommended: Add an outputSchema so models can better understand this tool's results." — the MCP server did not declare output schemas. Additionally, all tools were missing semantic operation hints (`readOnly, destructive, openWorld`), causing write and delete operations to be reported as read-only.

## Design
Added Zod output schemas to all 26 Swagger tools. Schemas are defined in type files (`portal-types.ts, registry-types.ts, user-management-types.ts`) and referenced via outputSchema on each tool definition.structuredContent is now set conditionally only for plain objects, while content is always populated — both fields coexist in the response rather than content being emptied out.
For tools that return arrays (`getTableOfContents, searchApis`) or raw strings (`getApiDefinition`), added `formatResponse `wrappers to wrap the result into an object that matches the declared output schema, enabling structured content to be set correctly.

## Changeset

- Added `outputSchema` to all 26 Swagger MCP tools

- Fixed `structuredContent` handling in SwaggerClient — `content` is always populated, `structuredContent` is added only for plain object responses

- Added `isError: true` to error responses

- Added `formatResponse` wrappers for getTableOfContents, searchApis, getApiDefinition, and deletePortalProduct to align array/string responses with declared schemas

- Introduced tool-constants.ts with `READ_ONLY, WRITE, WRITE_DESTRUCTIVE, WRITE_INTERNAL` presets and applied them to all 26 tools — correcting previously incorrect read-only annotations on write/delete operations

## Testing

Existing unit tests pass
Build passes with no TypeScript errors


src/swagger/client/tools.ts
```diff
@@ -152,6 +209,7 @@ export const TOOLS: SwaggerToolParams[] = [
       "Delete table of contents entry. Performs a soft-delete of an entry from the table of contents. Supports recursive deletion of nested items.",
     inputSchema: DeleteTableOfContentsArgsSchema,
     handler: "deleteTableOfContents",
+    ...WRITE_DESTRUCTIVE,
```
KacperSB, line 209: This one doesn't have `outputSchema` expected?
If there is none, maybe we should add it?

src/swagger/client/tools.ts
```diff
@@ -187,23 +251,34 @@ export const TOOLS: SwaggerToolParams[] = [
...
   },
   {
     title: "Create or Update API",
     toolset: "Registry API",
     summary:
       "Create a new API or update an existing API in SwaggerHub Registry for Swagger Studio. The API specification type (OpenAPI, AsyncAPI) is automatically detected from the definition content. APIs are always created with fixed values: version 1.0.0, private visibility, and automock disabled (these values cannot be changed). Returns HTTP 201 for creation, HTTP 200 for update. Response includes 'operation' field indicating whether it was a 'create' or 'update' operation along with API details and SwaggerHub URL.",
     inputSchema: CreateApiParamsSchema,
+    outputSchema: CreateApiOutputSchema,
     handler: "createOrUpdateApi",
+    ...WRITE_INTERNAL,
```
KacperSB: Just to be sure here, this tool writes to Registry, is that considered as `INTERNAL` write?

src/swagger/client/portal-types.ts
```diff
@@ -113,37 +113,54 @@ export const CreateTableOfContentsArgsSchema = z.object({
...
+    .meta({
+      // Structural rule (in addition to the .refine() below) so the generated
+      // JSON Schema states which fields are required per content.type, instead
+      // of leaving every field optional with the constraint only in prose.
+      allOf: [
+        {
+          if: { properties: { type: { const: "apiUrl" } } },
+          // biome-ignore lint/suspicious/noThenProperty: JSON Schema if/then/else keyword, not a thenable
+          then: { required: ["url"] },
+          message: "URL is required when content type is 'apiUrl'",
```
KacperSB, line 148: Not sure if that `message` is needed since later you use `.refine()` that passes correct message

KacperSB (approved):

src/swagger/client/portal-types.ts
bartosz-nowacki: Both are needed but serve different purposes. Msg in .meta() ends up in json schema sent to the map client, .refine() handles runtime Zod validation on our side.

src/swagger/client/tools.ts
bartosz-nowacki: Good catch, changed to `WRITE_DESTRUCITVE`. Previously, it implied openWorld: false, which means no external side effects, but this operation writes data to the registry. Corrected.

src/swagger/client/tools.ts
bartosz-nowacki: Good catch. The handler already returned structuredContent but the schema was missing, so MCP clients had no type information for the response.

src/swagger/client/portal-types.ts
KacperSB: I see, fair enough

lukasikp (approved):

lukasikp (approved):

PR #555, opened 2026-06-30
feat: Extend swagger_create_documentation_page to accept optional pageSlug param

lukasikp:
## Goal

Allow callers to control the URL slug of a documentation page instead of always having it auto-generated from the title.

## Design

Added pageSlug as an optional parameter. When provided it is passed directly to the table-of-contents API; when omitted the existing normalizeSlug(pageTitle) fallback runs unchanged. No existing behavior is affected.

## Changeset

Added optional pageSlug parameter to CreateDocumentationPageArgsSchema with format constraints documented in the description (3-255 chars, lowercase alphanumeric with hyphens/underscores/dots)
Updated createDocumentationPage handler to use pageSlug ?? normalizeSlug(pageTitle)
Updated swagger-portal-integration.md parameter table to include pageSlug
Updated CHANGELOG.md under Unreleased

## Testing

Added unit test: when pageSlug is provided. Tested locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- [Swagger] `create_documentation_page` now accepts an optional `slug` parameter. When provided, it is used as the page URL slug; otherwise the slug is generated from the page title.
+
 ## [0.27.1] - 2026-06-29
 - [Swagger]: Removed the https://swagger.mcp.smartbear.com/mcp streamable-http remote from the remotes array in server.json.[#550](https://github.com/SmartBear/smartbear-mcp/pull/550)
 ## [0.27.0] - 2026-06-29
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- [Swagger] `create_documentation_page` now accepts an optional `slug` parameter. When provided, it is used as the page URL slug; otherwise the slug is generated from the page title.
+- [Swagger] `create_documentation_page` now accepts an optional `pageSlug` parameter. When provided, it is used as the page URL slug; otherwise the slug is generated from the page title.
 - [Swagger] Added `list_suites` tool for discovering test suites in your Swagger Functional Testing workspace. Requires `SWAGGER_FUNCTIONAL_TESTING_API_TOKEN` env var.
 
 ### Fixed
--- a/src/swagger/client/api.ts
+++ b/src/swagger/client/api.ts
@@ -974,6 +974,7 @@ export class SwaggerAPI {
       source = "internal",
       order = 0,
       parentId = null,
+      pageSlug,
     } = args;
 
     if (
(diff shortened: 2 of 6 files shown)


szymon-pleszkun-sb (approved):

VarshaNagtilak1 (approved):

kacperdufrat (approved):

src/swagger/client/api.ts
```diff
@@ -999,13 +1000,13 @@ export class SwaggerAPI {
     }
     const section = sections.items[0];
 
-    const pageSlug = normalizeSlug(pageTitle);
+    const resolvedPageSlug = pageSlug ?? normalizeSlug(pageTitle);
```
KacperSB: If `pageSlug` is send as empty "", then it will be passed to `resolvedPageSlug` as the ?? will only trigger on `null` or `undefined`. 
I suggest to use `pageSlug || normalizeSlug(pageTitle)` instead to catch that

src/swagger/client/portal-types.ts
```diff
@@ -594,7 +594,13 @@ export const CreateDocumentationPageArgsSchema = z.object({
...
     .string()
     .describe(
-      "Title of the documentation page - will be displayed in navigation (3-255 characters, used to generate the page slug)",
+      "Title of the documentation page - will be displayed in navigation (3-255 characters)",
+    ),
+  pageSlug: z
+    .string()
+    .optional()
+    .describe(
+      "URL slug for the documentation page. 3-255 characters, lowercase, alphanumeric with hyphens, underscores, or dots (e.g. 'my-page'). If not provided, the slug is generated from the page title.",
```
KacperSB, line 603: This `zod` schema doesn't enforce any of described validation.
I suggest to add `.min(3).max(255)` - at minimum to validate length.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- [Swagger] `create_documentation_page` now accepts an optional `slug` parameter. When provided, it is used as the page URL slug; otherwise the slug is generated from the page title.
```
KacperSB: The parameter name is not `slug`, it's `pageSlug`

src/tests/unit/swagger/api.test.ts
```diff
@@ -966,6 +966,26 @@ describe("SwaggerAPI", () => {
       expect(result.pageDetails.slug).toBe("hello-world-123");
       expect(result.draftUrl).toContain(`/edit/content/${tocItemId}`);
     });
+
+    it("should use provided slug instead of generating one from title", async () => {
```
KacperSB, line 970: Some test that verifies if fallback works when `pageSlug` is `undefined` would be nice

KacperSB: And i guess 1 more that checks fallback when `pageSlug` is empty ""

KacperSB (changes requested):
Only requesting change as there is mistake in `CHANGELOG.md` which i think we should keep correct for outsiders

The rest are just suggestions

src/swagger/client/api.ts
lukasikp: good catch! done

src/swagger/client/portal-types.ts
lukasikp: Currently, none of the portal tools define min and max length in Zod, so I’ll hold off on that for now.

KacperSB (approved):
:shipit:

KrzysztofNocen (approved):

PR #563, opened 2026-07-03
Registry remove remote url workflow

szymon-pleszkun-sb:
## Goal

The registry's `validateNoDuplicateRemoteURLs` check rejects publishing a remote                                                                                                                           
  URL that is still claimed by any non-deleted version of another server. We want                                                                                                                            
  to publish `com.smartbear/swagger-mcp` with `https://swagger.mcp.smartbear.com/mcp`,                                                                                                                       
  but that URL is still referenced by older `com.smartbear/smartbear-mcp` versions                                                                                                                           
  (0.25.0–0.27.0), so the publish fails with a 400 conflict. This adds tooling to                                                                                                                            
  free the URL by removing it from those old versions.

## Design

The registry offers no "transfer URL" or "scope conflict check to latest"                                                                                                                                  
  operation — the only way a version stops holding a URL is to be `deleted` or                                                                                                                               
  edited. We chose **editing** (via `PUT /v0/servers/{name}/versions/{version}`)                                                                                                                             
  over deleting so the versions stay `active` and version history is preserved;                                                                                                                              
  the URL is simply removed from each version's `remotes`.                                                                                                                                                   
                                                                                                                                                                                                             
  Built as a generic, parameterized CLI (any server / any URL / any versions) so                                                                                                                             
  it's reusable for future remote-URL migrations, plus a `workflow_dispatch`                                                                                                                                 
  wrapper that reuses the same `github-oidc` + `dns` auth as                                                                                                                                                 
  `publish-swagger-mcp.yaml` (so no local DNS key is needed — it uses the                                                                                                                                    
  `GH_MCP_REGISTRY_KEY` secret). Defaults to a dry run; edits require an explicit                                                                                                                            
  `apply`. Edits are per-version and idempotent, so a partial failure is safe to                                                                                                                             
  re-run. 

## Changeset

- `scripts/registry-remove-remote.js` — CLI that auto-detects (or takes an                                                                                                                                 
    explicit list of) versions still holding a remote URL and removes it; dry-run                                                                                                                            
    by default, `--apply`/`--yes` to write, token from `mcp-publisher`.                                                                                                                                      
  - `.github/workflows/registry-remove-remote.yaml` — manual `workflow_dispatch`                                                                                                                             
    wrapper with `server` / `remote_url` / `versions` / `apply` inputs. 

## Testing

  - Dry run against the live registry correctly listed all 5 affected versions                                                                                                                               
    (0.25.0, 0.25.1, 0.26.0, 0.26.1, 0.27.0), preserving the bugsnag + zephyr                                                                                                                                
    remotes.                                                                                                                                                                                                 
  - Traced the edit path in the registry source: it only re-validates package                                                                                                                                
    ownership (`mcpName`), not remotes — and confirmed all 5 npm versions carry                                                                                                                              
    `mcpName: com.smartbear/smartbear-mcp`, so `--apply` will pass validation.


.github/workflows/publish-swagger-mcp.yaml
```diff
@@ -17,7 +17,7 @@ on:
 jobs:
   publish-swagger-mcp:
     name: Publish swagger-mcp to MCP Registry
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
```
KacperSB: Why such change?
I recommend to not use `-latest` as if malicious version releases this action will use it
When you point to specific version there is no such risk as the version releases are immutable

.github/workflows/registry-remove-remote.yaml
```diff
@@ -0,0 +1,73 @@
...
+          ./mcp-publisher login github-oidc
+          ./mcp-publisher login dns --domain smartbear.com --private-key "${{ secrets.GH_MCP_REGISTRY_KEY }}"
+
+      - name: Remove remote URL
+        run: |
+          node scripts/registry-remove-remote.js \
+            --server "${{ inputs.server }}" \
+            --remote-url "${{ inputs.remote_url }}" \
+            ${{ inputs.versions && format('--versions "{0}"', inputs.versions) || '' }} \
+            ${{ inputs.apply && '--apply --yes' || '' }}
```
KacperSB: I always recommend to use `env` vars instead of raw inputs to prevent script injection attacks [[source]](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable)

For this part you could change it to:

```yaml
  - name: Remove remote URL
    env:
      SERVER: ${{ inputs.server }}
      REMOTE_URL: ${{ inputs.remote_url }}
      VERSIONS: ${{ inputs.versions }}
      APPLY: ${{ inputs.apply }}
    run: |
      node scripts/registry-remove-remote.js \
        --server "$SERVER" \                                                                                                                                                                              
        --remote-url "$REMOTE_URL" \
        ${VERSIONS:+--versions "$VERSIONS"} \
        ${APPLY:+--apply --yes}
```

These conditionals at the end work like this:`${VAR:+word}` expands to `word` if `VAR` is set and non-empty, else nothing.

.github/workflows/registry-remove-remote.yaml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version-file: ".tool-versions"
+
+      - name: prepare mcp-publisher
+        run: |
+          curl -L "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/').tar.gz" | tar xz mcp-publisher
```
KacperSB: Here again, (if possible) i would recommend pinning to specific version tag instead of just latest

scripts/registry-remove-remote.js
```diff
@@ -0,0 +1,341 @@
...
+  console.log(`Mode     : ${args.apply ? "APPLY" : "dry-run"}\n`);
+
+  const allVersions = await fetchAllVersions(args.registry, args.server);
+  if (allVersions.length === 0) {
+    fail(`No versions found for ${args.server}`);
+  }
+
+  // Candidate = version whose server JSON still lists the remote URL.
+  let candidates = allVersions.filter((rec) =>
+    serverHasRemote(rec.server, args.remoteUrl),
```
KacperSB, line 261: This assumes `rec.server` is an object, are you sure it will always be?
Optionally you could add a guard for that

KacperSB (approved):
Some suggestions

KrzysztofNocen (approved):

PR #571, opened 2026-07-08
[Zephyr] TM4J-17879: Add a note in documentation about supported Zephyr addons.

PrzemyslawZnojekSB:
## Goal

Inform customers about supported Zephyr addons.

## Design

Typical change in documentation.

## Changeset

- `docs/products/SmartBear MCP Server/zephyr-integration.md`

## Testing
- Read the documentation.
- Read it again.
- Switched to light mode and read it again.
- Increased zoom to 125%, then decreased it to 90%, then settled on 100% after a rigorous calibration process.
- Tested while slightly hungry.
- Tested after coffee.


juro994 (approved):
🚀

PR #575, opened 2026-07-08
chore(pactflow): add pact tests for pactflow integration

kevinrvaz:
## Goal

- Add pact consumer tests for pactflow api integrations
- Add CI workflow for running and publishing pact contracts

## Changeset

Added pact tests to gate changes to pactflow


src/pactflow/tests/pactflow-ai.pact.test.ts
```diff
@@ -0,0 +1,138 @@
+/**
```
Ilia, line 1: can we move these into our own pactflow directory? 

I agree with @JP-Ellis  that we need to move product specific tests away from this shared directory

kevinrvaz: shifted to the pactflow folder, i dont see any comments from @JP-Ellis ?

prateek0411999: I think we should do it for CI scripts as well

kevinrvaz: the CI script would be common for others(bugsnag, etc) for them to adopt pact tests for their integrations

tokirhusen-smartbear (approved):

PR #582, opened 2026-07-09
chore: shift all unittests into respective product folders

kevinrvaz:
## Goal

Shifted all unit tests from the root level tests folder to individual product folders based on internal discussion with team.


JP-Ellis (changes requested):
I would much prefer adopintg a simpler naming convention, which co-locates tests with the files. E.g.:

- `path/to/file.ts`
- `path/to/file.test.ts`

But in general, I very much agree with this change. As this project grows, we want teams to own their sections, and having tests and code spread out just makes that more difficult.

src/bearq/tool/environments/list-environments.test.ts
```diff
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import createFetchMock from "vitest-fetch-mock";
-import { ListEnvironments } from "../../../../../bearq/tool/environments/list-environments";
+import { ListEnvironments } from "./list-environments";
```
JP-Ellis, line 3: That's _much_ nicer 😆

vitest.config.ts
```diff
@@ -54,8 +54,8 @@ export default defineConfig({
       // Include source map support
       reportsDirectory: "./coverage",
     },
-    include: ["src/tests/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}"],
-    exclude: ["src/pactflow/tests/**"],
+    include: ["src/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}"],
```
JP-Ellis: Nit:


```suggestion
    include: ["**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}"],
```

vitest.pact.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
     globals: true,
     environment: "node",
     include: [
-      "src/pactflow/tests/**/*.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}",
+      "src/pactflow/**/*.pact.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}",
```
JP-Ellis: Similar nit:

```suggestion
      "**/*.pact.{test,spec}.{js,mjs,cjs,ts,mts,cts,jsx,tsx}",
```

JP-Ellis (approved):
Couple of nits. Since we have now one main `src/` directory, it becomes a bit redundant to scope to `src/**/*.test.ts`, when we could just match any `**/*.test.ts`. Looks so much better though!

JP-Ellis (approved):

PR #583, opened 2026-07-10
fix(pactflow): update AI pact interactions to expect 202 async response

kevinrvaz:
## Goal

Correcting response code for pactflow-ai cdct tests which return a 202


Saup21 (approved):
LGTM

PR #587, opened 2026-07-13
[Reflect] RF-5198: Add new create_test and create_segment MCP tools

tmcneal:
## Goal

The primary use-case for these endpoints are to import tests from other tools such as open source frameworks (Selenium, Cypress, Playwright) or other SmartBear products (TestComplete, BearQ).

## Design

The tool definitions provide guidance on how to construct a test or segment, and the API endpoints backing these tools provide feedback when an agent sends an invalid request.

## Changeset

Add two new Reflect tools: `create_test` and `create_segment`

## Testing
- Took a few Playwright tests and verified the agent could create equivalent web tests in Reflect
- Took a few OpenAPI specs and verified the agent could create API tests that exercised the endpoints documented in the OpenAPI specs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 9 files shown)


mahavir1166 (approved):
🚢

mahavir1166 (approved):

PR #596, opened 2026-07-15
chore(deps): bump biome

JP-Ellis:
## Summary

Strengthen linting across the board to catch low-lying issues, and resolve the new lints.

## Background

Biome is an excellent and fast linter, and in many cases, has automatic fixes which are safe to apply. We had been using biome, but I have opted to strengthen the linting rules significantly.

A lot of the fixes were automatic, with some requiring manual intervention. Also reviewed manually. I would recommend reviewing the changes on a commit-by-commit bases to disambiguate automatic fixes, from manual ones.


src/bearq/tool/tasks/chat-with-qa-lead.ts
```diff
@@ -1,9 +1,9 @@
 import type { ToolCallback } from "@modelcontextprotocol/sdk/server/mcp.js";
 import type { ZodRawShape } from "zod";
 import { z } from "zod";
-import { Tool, ToolError } from "../../../common/tools";
-import type { ToolParams } from "../../../common/types";
-import type { BearQClient } from "../../client";
+import { Tool, ToolError } from "../../../common/tools.ts";
```
kevinrvaz, line 4: we have config which allows the import to work without the .ts, and that can reduce some of the diff

PR #609, opened 2026-07-22
[Zephyr] Refactor to keep REST API Schema unmodified and enforce type validation

assisblf:
**TL;DR:** `deepMerge` now coerces its output to a target Zod schema (dropping unknown keys at any depth), and the test-case/test-cycle update tools expose `folder`/`component`/`owner`/`jiraProjectVersion` as plain IDs instead of nested objects (just as before this change, but now in a different place).

## Goal

Update tools were manually deleting specific fields (`createdOn`, `links`, ...) after merging updates into the existing resource, which is brittle against new read-only API fields. Also, the REST API models `folder`/`component`/`owner`/`jiraProjectVersion` as `{ id }` / `{ accountId }` objects, which not all LLMs handle reliably as tool arguments.

## Design

- `deepMerge` takes a target Zod schema and parses the merged result through a deep-stripped version of it (new `deepStrip` helper), dropping any undeclared key at any nesting depth instead of relying on manual `delete` calls.
- `update-test-case.ts` / `update-test-cycle.ts` expose `folder`/`component`/`owner`/`jiraProjectVersion` as plain IDs in their tool input, expanding them back into the real `{ id }` / `{ accountId }` shape before merging/PUTing.
- Dropped now-unnecessary null special-casing for `plannedStartDate`/`plannedEndDate` (API already no-ops on null there).

## Changeset

- `src/zephyr/common/utils.ts` — schema-aware `deepMerge` + new `deepStrip`.
- `src/zephyr/common/rest-api-schemas.ts` — regenerated from the OpenAPI spec.
- `src/zephyr/tool/test-case/update-test-case.ts`, `.../test-cycle/update-test-cycle.ts` — primitive ID tool inputs, schema-aware merge.
- `src/common/__snapshots__/server-definitions.test.ts.snap` — regenerated for updated tool examples.

## Testing

- `vitest run` — full suite passing (153 files / 2384 tests).
- New tests for `deepMerge`/`deepStrip` (schema stripping at depth, leaf validation, `default()`/`union` cases, undefined-updates short-circuit) and for folder/owner/component ID expansion in the update tools.
- Tested manually by talking to ChatGPT to get test cases, create folders, and update test cases
- Tested manually with Remove MCP against Rovo (using Gemini model) with the same scenarios that I used with ChatGPT

### DIFF
--- a/src/zephyr/common/rest-api-schemas.ts
+++ b/src/zephyr/common/rest-api-schemas.ts
@@ -21,18 +21,36 @@ For accessing the API, you must generate an access key in Jira. To generate an a
 choose the option “Zephyr API keys". For more information, please check out the [documentation](https://support.smartbear.com/zephyr/docs/en/rest-api/api-access-tokens-management.html).
 
 ## Accessing the API
-The API is available at the following base URL:
+The API is available at different base URLs depending on your data residency region:
+
+**US region:**
 ```
 https://api.zephyrscale.smartbear.com/v2
 ```
-For example, the final URL for retrieving test cases would be:
-```
-https://api.zephyrscale.smartbear.com/v2/testcases
-```
-For EU region, the API can be accessed at the following base URL:
+
+**EU(eu-west-1) region:**
 ```
 https://eu.api.zephyrscale.smartbear.com/v2
 ```
+
+**AU(ap-southeast-1) region:**
+```
+https://au.api.zephyrscale.smartbear.com/v2
+```
+
+**DE(eu-central-1) region:**
+```
+https://de.api.zephyrscale.smartbear.com/v2
+```
+
+### Example API Calls
+
+Append the specific endpoint path to your region's base URL. For example, to retrieve test cases:
+
+- **US:** `https://api.zephyrscale.smartbear.com/v2/testcases`
+- **EU:** `https://eu.api.zephyrscale.smartbear.com/v2/testcases`
+- **AU:** `https://au.api.zephyrscale.smartbear.com/v2/testcases`
+- **DE:** `https://de.api.zephyrscale.smartbear.com/v2/testcases`
 
 ## Making Authenticated Requests
 To authenticate subsequent API requests, you must provide a valid token in an HTTP header, which is the key generated on the previous step:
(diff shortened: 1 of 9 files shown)


src/zephyr/common/rest-api-schemas.ts
```diff
@@ -21,19 +21,37 @@ For accessing the API, you must generate an access key in Jira. To generate an a
...
+https://de.api.zephyrscale.smartbear.com/v2
 ```
 
+### Example API Calls
+
+Append the specific endpoint path to your region's base URL. For example, to retrieve test cases:
+
+- **US:** `https://api.zephyrscale.smartbear.com/v2/testcases`
+- **EU:** `https://eu.api.zephyrscale.smartbear.com/v2/testcases`
+- **AU:** `https://au.api.zephyrscale.smartbear.com/v2/testcases`
```
vmrvictor, line 52: 👍

src/zephyr/common/rest-api-schemas.ts
```diff
@@ -484,150 +580,385 @@ export const CreateTestCase201Response = zod
...
+export const listTestCasesCursorPaginatedQueryStartAtIdDefault = 0;
+export const listTestCasesCursorPaginatedQueryStartAtIdMin = 0;
+
+export const ListTestCasesCursorPaginatedQueryParams = zod.object({
+  projectKey: zod
+    .string()
+    .regex(listTestCasesCursorPaginatedQueryProjectKeyRegExp)
+    .optional()
+    .describe("Jira project key filter"),
+  folderId: zod.number().min(1).optional().describe("Folder ID filter"),
```
vmrvictor, line 867: 👍

src/zephyr/common/rest-api-schemas.ts
```diff
@@ -1204,18 +1683,34 @@ export const UpdateTestCaseBody = zod
...
       .describe("ID and link to the status resource."),
     folder: zod
-      .number()
-      .min(1)
-      .nullable()
-      .optional()
-      .describe(
-        "The ID of the folder, to remove folder set it's value to null",
-      ),
+      .object({
```
vmrvictor, line 1686: so this folder containing self and id that is nullish will not be an issue?

src/zephyr/common/rest-api-schemas.ts
```diff
@@ -1204,18 +1683,34 @@ export const UpdateTestCaseBody = zod
...
+      })
+      .strict()
+      .nullish()
+      .describe("ID and link to the folder resource."),
     owner: zod
-      .string()
-      .regex(updateTestCaseBodyOwnerAccountIdRegExp)
-      .nullable()
-      .describe("Atlassian Account ID of the Jira user."),
+      .object({
```
vmrvictor, line 1698: the same for owner? or this is being solved at another level with the new tools?

src/zephyr/common/utils.ts
```diff
@@ -1,45 +1,111 @@
...
-    !Array.isArray(objectWithUpdates[key]) &&
-    baseObject[key] &&
-    typeof baseObject[key] === "object" &&
-    !Array.isArray(baseObject[key])
-  );
+function stripNode(schema: z.ZodTypeAny): z.ZodTypeAny {
+  // biome-ignore lint/suspicious/noExplicitAny: traversing zod's internal defs
+  const def = schema.def as any;
+
+  switch (def.type) {
```
vmrvictor, line 84: can you explain what is happening here pls?

src/zephyr/common/utils.ts
```diff
@@ -1,45 +1,111 @@
...
- * Deep merge two objects, with properties from 'updates' taking precedence.
- * For nested objects, this merges rather than replaces them entirely.
+ * Deep merge two objects and coerce the result to a target Zod schema.
+ *
+ * Merge rules:
+ * - Values from `objectWithUpdates` take precedence over `baseObject`.
+ * - `undefined` in `objectWithUpdates` is skipped (the base value is kept).
+ * - `null` is treated as a real value and overwrites the base value.
+ * - Plain objects are merged recursively; arrays and non-plain objects
+ *   (e.g. Date, Map, class instances) are replaced wholesale.
```
juro994: What does the _are replaced **wholesale**_ mean? I see this term also in tests.

Can we change it to "as a whole"?

JP-Ellis: The adjective _wholesale_ in this context means 'involving everyone or everything; complete'. So it is correct, both in its usage and placement.

src/zephyr/common/rest-api-schemas.ts
assisblf: Just to make it clear, this isn't different from the previous schema. This relates to the query params, so it was always an optional number

src/zephyr/common/rest-api-schemas.ts
assisblf: No, because this is the REST API schema. We're adapting this for LLMs to understand the schema at the tool input schema (see **[here](https://github.com/SmartBear/smartbear-mcp/pull/609/changes#diff-0f84e4191eaa60effd58dd1b6506a8005ff0261e7537b25ae6acc883de69fb3cR20)** and **[here](https://github.com/SmartBear/smartbear-mcp/pull/609/changes#diff-0f84e4191eaa60effd58dd1b6506a8005ff0261e7537b25ae6acc883de69fb3cR45)**)

src/zephyr/tool/test-case/update-test-case.ts
```diff
@@ -1,15 +1,39 @@
...
   UpdateTestCaseParams,
+  updateTestCaseBodyOwnerAccountIdRegExp,
 } from "../../common/rest-api-schemas";
 import { deepMerge } from "../../common/utils";
 
+// The update API expects a folder/component/owner as `{ id }` / `{ accountId }`
+// optional objects, but that optional objects are not supported the same way by all LLMs.
+// Expose them as plain IDs here and expand them back into objects before merging into the
+// real UpdateTestCaseBody shape (i.e., the one required by the REST API).
+const UpdateTestCaseBodyToolInput = UpdateTestCaseBody.extend({
```
vmrvictor, line 19: ok here we have the magic so we take the rest api but overwrite the things not supported folder/component/owner

juro994 (approved):

vmrvictor (approved):

src/zephyr/common/rest-api-schemas.ts
assisblf: Same as discussed **[here](https://github.com/SmartBear/smartbear-mcp/pull/609#discussion_r3637545360)** (I know we talked about it on the call, I'm just writting for other readers)

src/zephyr/common/utils.ts
assisblf: In short, this method receives a Zod schema that may contain ".strict()" attributes and transforms them to the **default** Zod strictness.

For more context:
* `z.object({...})` → strips unknown keys (the v4 default)
    * This is what we want when merging 2 objects with a desired target schema. We just want to strip out properties that we don't want from GET requests, for example.
* `z.strictObject(...) / .strict()` → rejects unknown keys
    * This is what we want to do in arguments coming from LLM. We want to force them to understand they provided something wrong instead of silently failing.
* `z.looseObject(...) / .passthrough()` → keeps unknown keys
    * we don't want this for most cases

src/zephyr/common/utils.ts
assisblf: Thanks for explaining @JP-Ellis and thanks @juro994 for recommending the improvement too

I just reworded it to avoid confusion moving forward

Addressed on 2cb3e079966932eec12d6c8ddf960c71ee2ea2f5

vmrvictor (approved):

PR #617, opened 2026-07-28
Fix session isolation to prevent credential and base path leakage

tokirhusen-smartbear:
Fixes [AIA-473](https://smartbear.atlassian.net/browse/AIA-473), [AIA-472](https://smartbear.atlassian.net/browse/AIA-472) two security vulnerabilities where shared singleton client state could leak between concurrent sessions:

1. **Base Path Injection / SSRF**: a malicious session could inject a custom `Swagger-Portal-Base-Path` into the shared `SwaggerClient` singleton, causing a victim's subsequent API calls to be sent to an attacker-controlled host.

2. **Credential Leakage**: an unauthenticated session initialized after an authenticated one would reuse the singleton's `this.api` and `this._apiKey`, giving the attacker access to the victim's Swagger tools and credentials.

`ClientRegistry.configure()` now calls `cloneClient()` to produce a fresh client instance per session. Each session's state is fully isolated — base paths and tokens set during one session cannot affect any other.

Changed Swagger base-path config fields from `z.string()` to `z.url()` so non-URL values are rejected at parse time and `MCP_ALLOWED_ENDPOINTS` allowlist enforcement applies to them.

**Tested**
- Unauthenticated session has zero Swagger tools (no credential reuse)
- Two concurrent sessions receive independent client clones with separate state (verified by unit test and live curl tests)

[AIA-473]: https://smartbear.atlassian.net/browse/AIA-473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AIA-472]: https://smartbear.atlassian.net/browse/AIA-472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Zephyr] Refactor to keep REST API schemas unmodified and ensure correct payload for update operations [#609](https://github.com/SmartBear/smartbear-mcp/pull/609)
 
+### Security
+
+- Fixed session isolation vulnerability (Base Path Injection / SSRF): `ClientRegistry.configure()` now calls `cloneClient()` to produce a fresh client instance per session, preventing per-session state (base-path URLs, auth tokens) stored on one session's client from leaking into concurrent sessions.
+- Fixed credential leakage vulnerability: because each session receives its own client clone, an unauthenticated session no longer inherits a previously authenticated session's `SwaggerClient` API instance or API key. An unauthenticated session has no Swagger tools available.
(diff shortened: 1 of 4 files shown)


krishna-smartbear (approved):
Approved

romin21 (approved):

dev-smartbear (approved):

abd-smartbear (approved):

PR #618, opened 2026-07-28
feat: added url in standardize-api tool when new version is saved

VarshaNagtilak1:
## Goal
SWG-21072:
We want to url should be return by tool so Agent will not hallucinate it or we  don't need to hardcode it in skill base on env.

## Design

to avoid hallucination and env specific url generation

## Changeset
If the fixed definition was saved, the response includes a `url` to view the API in SwaggerHub (server returns `savedVersion`); otherwise `url` is omitted

## Testing

standardize api tool call if url present in response or not.

### DIFF
--- a/src/swagger/client/tools.ts
+++ b/src/swagger/client/tools.ts
@@ -323,7 +323,7 @@ export const TOOLS: SwaggerToolParams[] = [
     title: "Standardize API",
     toolset: "Registry API",
     summary:
-      "Standardize and fix an API definition using AI to ensure compliance with governance policies. Scans the API definition for standardization errors and automatically fixes them using SmartBear AI. Optionally provide 'newVersion' (e.g. patch bump '1.0.0' → '1.0.1') to save the fixed definition as a new version — omitting it will overwrite the current version. Returns the number of errors found, the fixed definition, and a URL to view the standardized API (URL is environment-specific and determined by the server configuration). Use this tool when users ask to standardize, fix, govern, or ensure governance compliance of APIs.",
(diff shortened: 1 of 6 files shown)


kacperdufrat (approved):

src/swagger/client/tools.ts
```diff
@@ -322,7 +322,7 @@ export const TOOLS: SwaggerToolParams[] = [
...
     summary:
-      "Standardize and fix an API definition using AI to ensure compliance with governance policies. Scans the API definition for standardization errors and automatically fixes them using SmartBear AI. Optionally provide 'newVersion' (e.g. patch bump '1.0.0' → '1.0.1') to save the fixed definition as a new version — omitting it will overwrite the current version. Returns the number of errors found and the fixed definition if successful. Use this tool when users ask to standardize, fix, govern, or ensure governance compliance of APIs.",
+      "Standardize and fix an API definition using AI to ensure compliance with governance policies. Scans the API definition for standardization errors and automatically fixes them using SmartBear AI. Optionally provide 'newVersion' (e.g. patch bump '1.0.0' → '1.0.1') to save the fixed definition as a new version — omitting it will overwrite the current version. Returns the number of errors found, the fixed definition, and a URL to view the standardized API (URL is environment-specific and determined by the server configuration). Use this tool when users ask to standardize, fix, govern, or ensure governance compliance of APIs.",
```
lukasikp:
```suggestion
      "Standardize and fix an API definition using AI to ensure compliance with governance policies. Scans the API definition for standardization errors and automatically fixes them using SmartBear AI. Optionally provide 'newVersion' (e.g. patch bump '1.0.0' → '1.0.1') to save the fixed definition as a new version — omitting it will overwrite the current version. Returns the number of errors found, the fixed definition, and a URL to view the standardized API. Use this tool when users ask to standardize, fix, govern, or ensure governance compliance of APIs.",
```
Minor change suggestion. This message is redundant and unnecessarily adds context to the LLM.

lukasikp (approved):

lukasikp (approved):

PR #624, opened 2026-07-29
feat: update portal tool to create proper subdomain

kacperdufrat:
## Goal

The `resolve_organization_portal` tool created the portal subdomain from the organization name only. Now, it requires the same rule as the Portal UI: slugify the organization name and always add a random suffix (hyphen + 3 alphanumeric characters), so the subdomain is unique from the start.

## Design

The subdomain logic follows the Portal UI convention. The organization name is turned into a valid slug (lowercase, only a-z0-9-, 3-20 characters) and a random 3-character suffix is appended, even when there is no collision. On a 409 conflict, creation is retried with a fresh suffix.  For `create_portal`, per the refinement decision, the convention is only suggested in the parameter description and the MCP client still picks the name.

## Changeset

`portal-utils.ts`: added convertToValidSubdomain and appendRandomSuffix (ported from the Portal UI), reworked buildSubdomainCandidate to combine them, removed the old slugify and buildSuffixedSubdomain helpers.
`api.ts`: createPortalForOrganization now builds candidates with buildSubdomainCandidate and retries with a fresh random suffix on conflict.
`portal-types.ts`: the create_portal subdomain description now recommends the convention (slug + random 3-character suffix) with an example.
Updated the server-definitions snapshot.

## Testing

New unit tests for the subdomain helpers in portal-utils.test.ts: slugification, padding of short names, length limits, suffix padding and the hyphen-at-cut-point edge case.
Updated resolveOrganizationPortal tests in api.test.ts with a mocked Math.random for deterministic suffixes, including the retry-on-conflict path.


src/swagger/client/portal-utils.ts
```diff
@@ -21,61 +23,61 @@ export function isLengthWithin(
...
 /**
- * Sanitize a free-form string into a subdomain-safe slug: lowercase,
- * alphanumeric segments separated by single hyphens, trimmed to maxLength
- * with no trailing hyphen. Matches ^[a-z0-9]+(-[a-z0-9]+)*$.
+ * Sanitize a free-form string into a subdomain-safe slug.
+ * Matches ^[a-z0-9]+(-[a-z0-9]+)*$ within the 3-20 length limits.
  */
-export function slugify(value: string, maxLength: number): string {
-  return value
+export function convertToValidSubdomain(
```
KacperSB, line 29: This function now strip spaces, but before this change the `slugify` would replace spaces with hyphens.
Just wanted to confirm - is this OK for the Portal UI?

src/swagger/client/portal-utils.ts
```diff
@@ -21,61 +23,61 @@ export function isLengthWithin(
...
+ * Append a random "-<suffix>" to a subdomain base while staying within the
+ * maximum length. Each call yields a fresh candidate, used to retry portal
+ * creation after a subdomain conflict. The suffix is padded to a full 3
+ * characters and no hyphen is left at the cut point; the base must already be
+ * a sanitized slug (see convertToValidSubdomain) for the result to be valid.
  */
-export function buildSubdomainCandidate(
-  organizationId: string,
-  organizationName?: string,
+export function appendRandomSuffix(
```
KacperSB, line 58: Since this function (as per doc above) requires a sanitized slug (`base` in fact) for the result to be valid, then i would recommend keeping it not exported - so that it won't be misused by somebody n the future. Right now this code only needs `buildSubdomainCandidate` exported as this method is used only there.

Or - if you want to keep it - an assertion for the suffix would be nice.

src/swagger/api.test.ts
```diff
@@ -1082,6 +1082,18 @@ describe("SwaggerAPI", () => {
     const organizationId = "d36ff595-6406-4a1b-a1d5-90153ac8f334";
     const jsonHeaders = { "content-type": "application/json" };
 
+    const suffixRandom = 0.123456789;
+    const suffixFor = (r: number) =>
+      r.toString(36).substring(2, 5).padEnd(3, "0");
+
+    beforeEach(() => {
+      vi.spyOn(Math, "random").mockReturnValue(suffixRandom);
```
KacperSB, line 1090: [optional] This adds spy for every test in `resolveOrganizationPortal`, even for ones that do not call it - this could be scoped more narrow for tests that in fact use it.

src/swagger/client/api.ts
```diff
@@ -460,19 +459,15 @@ export class SwaggerAPI {
   /**
    * Create a portal for an organization that has none yet. The subdomain is
    * derived from the organization name; on a subdomain conflict a candidate
-   * suffixed with part of the organization UUID is retried.
+   * with a fresh random suffix is retried.
```
KacperSB: [optional] This could be misleading as if i understand correctly the suffix is always appended (even without a conflict), if that's true - maybe we can do something like:
```ts
/**
* Create a portal for an organization that has none yet. The subdomain is
* derived from the organization name with additional random suffix.
*/
```

KacperSB (approved):
lgtm, few suggestions

src/swagger/client/portal-utils.ts
kacperdufrat: Yes, intentional. It was decided on SWG-21041 to reuse the Portal UI logic as-is (convertToValidSubdomain in swaggerhub-portal-frontend strips spaces too), so MCP now produces the same subdomains as the UI.

src/swagger/client/portal-utils.ts
kacperdufrat: It's exported on purpose for direct unit testing, portal-utils.test.ts covers its edge cases separately. Same pattern as in swaggerhub-portal-frontend, where getUniqueSubdomain is also exported and tested directly.

src/swagger/api.test.ts
kacperdufrat: 6 of 8 tests here go through the creation path that uses Math.random, so the shared spy keeps them all deterministic. I'd keep it to avoid repeating the setup.

src/swagger/client/api.ts
kacperdufrat: Good catch, reworded as suggested. I only kept a note that the suffix logic mirrors the Portal UI.

lukasikp (approved):

lukasikp (approved):

PR #627, opened 2026-07-30
fix: swagger mcp tool open world setting

kacperdufrat:
## Goal

Fix wrong tool annotations. Twelve Swagger MCP tools were marked `Open World: True`, but they only work on data inside Swagger, not external systems. This was flagged during the MCP server submission (SWG-21378).

## Design

The `openWorld: true` flag came from the shared `WRITE` and `WRITE_DESTRUCTIVE` presets, so every write tool got it. Changed the default in both presets to false. Only the two AI-based tools (`create_api_from_prompt`, `standardize_api`) still need true, so we set `openWorld: true` inline on just those two.

## Changeset

`tool-constants.ts` - set `openWorld: false` in `WRITE` and `WRITE_DESTRUCTIVE`.
`tools.ts` - set `openWorld: true` inline on "Create API from Prompt" and "Standardize API".

## Testing

No new tests needed - adjusted snapshot.


lukasikp (approved):

KacperSB (approved):
:shipit:

lukasikp (approved):

PR #629, opened 2026-08-03
Migration to sdk-v2/codemod v1-to-v2 changes applied

dev-smartbear:
Migrating to sdkv2 modelcontext protocol.

Goal
Migrate the MCP server from the old SDK (@modelcontextprotocol/sdk, v1) to the new v2 beta SDK, since v1 is being deprecated.

Design
The codemod did the bulk of the mechanical rewrite but left ~85 files broken (wrong handler types) and a handful of spots it flagged for manual review (context object shape changed in v2). Fixed those directly and kept scope tight — deferred the bigger transport rewrite (createMcpHandler) to a separate task.

Changeset
Swapped SDK dependency to the v2 packages.
Fixed handler/prompt types across all tool and prompt files.
Updated common/server.ts to match v2's schema and context handling.
Fixed the couple of spots needing manual context-object renames (extra → ctx).
Updated tests to match.
Rebased onto latest main, resolving conflicts with unrelated changes that landed there in the meantime.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "@smartbear/mcp",
   "version": "0.33.0",
+  "engines": {
+    "node": ">=20"
+  },
   "description": "MCP server for interacting SmartBear Products",
   "keywords": [
     "smartbear",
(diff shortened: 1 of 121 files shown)


package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "@smartbear/mcp",
   "version": "0.33.0",
+  "engines": {
+    "node": ">=20"
```
kevinrvaz, line 5: smartbear-mcp runs on node 22 not node 20, increase this

PR #630, opened 2026-08-03
fix(functional-testing): update Create Suite tool to match latest API contract

bartoszkosicki13:
## Goal

  The Reflect API backing the Create Suite tool changed its contract: the `runApiTestsBlocks`
  field was renamed to `runApiTests`, it is now required (the API rejects a missing or empty
  list with a 400 — creating an "empty" suite with no tests is no longer supported), and
  `maxRetryAttempts` now accepts 0-3 instead of 1-3 (0/omit = no retry). The MCP tool needed
  to be updated to match, otherwise every `create_suite` call would fail against the current API.

  ## Design

  - Renamed `runApiTestsBlocks` → `runApiTests` end-to-end (zod schema, API request body, tests)
    to mirror the API's field name exactly.
  - Made `runApiTests` required with `.min(1)` in the zod schema, and updated the tool's
    `description`/`summary` text to make this obvious to the calling LLM — otherwise it could
    treat the field as optional configuration and omit it, wasting a round-trip on a 400.
  - Widened `maxRetryAttempts` from `min(1)` to `min(0)` and updated its description.
  - Considered adding a `.refine()` on the top-level schema to enforce unique block `title`s
    client-side, but reverted it: this repo's `schemaToRawShape` (`src/common/server.ts`) only
    unwraps `ZodObject`/`ZodIntersection`, not `ZodEffects` produced by `.refine()` — wrapping
    the schema would have silently dropped the *entire* input schema during tool registration.
    Left the uniqueness constraint as a documented rule instead; the server already returns a
    clear 400 for duplicate titles.
  - Added a short cross-reference from `create_test`'s description pointing to `create_suite`,
    so the LLM discovers that created tests can be grouped into a Suite.
  - Filled in a "Suite Creation" section in the SFT product doc, which was missing entirely for
    this tool, and added a CHANGELOG entry.

  ## Changeset
  
  - `src/swagger/client/functional-testing-types.ts`: renamed field, required `runApiTests`,
    `maxRetryAttempts` range 0-3.
  - `src/swagger/client/functional-testing-api.ts`: `createSuite()` sends `runApiTests`.
  - `src/swagger/client/functional-testing-tools.ts`: updated `Create Suite` summary and
    `Create Test` summary (cross-reference).
  - `docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md`: added
    `create_suite` documentation section.
  - `CHANGELOG.md`: added Unreleased entry.
  - `functional-testing-api.test.ts`, `swagger-client-ft.test.ts`: updated to the renamed/required
    field; removed the now-invalid "creates an empty suite" test case.

  ## Testing
  
  - `npx tsc --noEmit` — no errors.
  - `npx vitest run src/swagger/functional-testing` — 107 tests passing.
  - `npx biome lint` / `npx biome format --write` — clean on all changed files.

### DIFF
--- a/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
+++ b/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
@@ -40,7 +40,12 @@ All tools listed below are only available through the Local MCP Server. They are
 
 #### `create_suite`
 
-- Purpose: Creates a new test Suite in your Swagger Functional Testing workspace. Use this tool when you need to group existing tests into a Suite for collective execution. Requires a `name` and `runApiTests`, one or more ordered blocks of tests to run — each block requires a non-empty array of `testIds` (from `list_tests`) and may set `parallel` (run the block's tests in parallel instead of sequentially, default `false`), `maxRetryAttempts` (0-3, retry a block's failed tests before they count as failed, default no retry), and a `title` (must be unique across the Suite's blocks). Blocks always run one after another. Optionally accepts `agentName` to save a tunnel agent override for future runs of the Suite.
+- Purpose: Creates a new test Suite in your Swagger Functional Testing workspace by grouping existing tests into ordered blocks for collective execution. Requires a `name` and one or more `runApiTests` blocks, each with a non-empty `testIds` array (from `list_tests`). Blocks always run one after another; within a block, tests run sequentially by default. Each block may optionally set:
+  - `parallel` — run the block's tests in parallel instead of sequentially (default `false`)
+  - `maxRetryAttempts` — retry a block's failed tests before they count as failed, 0-3 (default no retry)
+  - `title` — a label that must be unique across the Suite's blocks
+
+  Optionally accepts `agentName` to save a tunnel agent override for future runs of the Suite.
 - Returns: The created Suite's `id`, `slug`, and `url`.
 - Use case: Group existing tests into a Suite, optionally with parallel/sequential blocks and retry behavior, for collective execution.
 
--- a/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
+++ b/docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
@@ -36,6 +36,16 @@ All tools listed below are only available through the Local MCP Server. They are
 
 ---
 
+### Suite Creation
+
+#### `create_suite`
+
+- Purpose: Creates a new test Suite in your Swagger Functional Testing workspace. Use this tool when you need to group existing tests into a Suite for collective execution. Requires a `name` and `runApiTests`, one or more ordered blocks of tests to run — each block requires a non-empty array of `testIds` (from `list_tests`) and may set `parallel` (run the block's tests in parallel instead of sequentially, default `false`), `maxRetryAttempts` (0-3, retry a block's failed tests before they count as failed, default no retry), and a `title` (must be unique across the Suite's blocks). Blocks always run one after another. Optionally accepts `agentName` to save a tunnel agent override for future runs of the Suite.
+- Returns: The created Suite's `id`, `slug`, and `url`.
+- Use case: Group existing tests into a Suite, optionally with parallel/sequential blocks and retry behavior, for collective execution.
+
+---
+
 ### Test Execution
 
 #### `run_test`
--- a/src/swagger/client/functional-testing-types.ts
+++ b/src/swagger/client/functional-testing-types.ts
@@ -69,36 +69,60 @@ export const RunApiTestsBlockSchema = z.object({
     ),
 });
 
-export const CreateFunctionalTestingSuiteParamsSchema = z.object({
-  name: z.string().describe("Name for the new suite").trim().min(1),
-  agentName: z
-    .string()
-    .trim()
-    .min(1)
-    .optional()
-    .describe(
-      "Tunnel agent name to save as this suite's tunnel override for future runs.",
-    ),
-  runApiTests: z
-    .array(RunApiTestsBlockSchema)
-    .min(1)
-    .describe(
-      'Required — ordered groups ("blocks") of tests to run one after another. ' +
-        "Must include at least one entry; suites cannot be created without a workflow. " +
-        "Within a block, tests run sequentially unless `parallel` is set. " +
-        "Block `title`s must be unique within the suite.",
-    ),
-});
+export const CreateFunctionalTestingSuiteParamsSchema = z
+  .object({
+    name: z.string().describe("Name for the new suite").trim().min(1),
+    agentName: z
+      .string()
+      .trim()
+      .min(1)
+      .optional()
+      .describe(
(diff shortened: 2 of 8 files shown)


docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
```diff
@@ -36,6 +36,16 @@ All tools listed below are only available through the Local MCP Server. They are
 
 ---
 
+### Suite Creation
```
ernest-bednarczyk, line 39: Just a thought as it is pattern we had already. I wonder if we should rather group those docs entries by entity rather than type of the action. 🤔 
Test tools:
...
Suite Tools:
...

ernest-bednarczyk (approved):
All good. I like `superRefine` usage and good test coverage.

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
```diff
@@ -36,6 +36,16 @@ All tools listed below are only available through the Local MCP Server. They are
...
+
+#### `create_suite`
+
+- Purpose: Creates a new test Suite in your Swagger Functional Testing workspace. Use this tool when you need to group existing tests into a Suite for collective execution. Requires a `name` and `runApiTests`, one or more ordered blocks of tests to run — each block requires a non-empty array of `testIds` (from `list_tests`) and may set `parallel` (run the block's tests in parallel instead of sequentially, default `false`), `maxRetryAttempts` (0-3, retry a block's failed tests before they count as failed, default no retry), and a `title` (must be unique across the Suite's blocks). Blocks always run one after another. Optionally accepts `agentName` to save a tunnel agent override for future runs of the Suite.
+- Returns: The created Suite's `id`, `slug`, and `url`.
+- Use case: Group existing tests into a Suite, optionally with parallel/sequential blocks and retry behavior, for collective execution.
```
KacperSB: IMO this is hard to read a little, let's remember that this md file will serve a documentation that is meant to be read by clients, not necessary LLMs

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
KacperSB: I was also thinking about that, but let's maybe leave that for other PR
In my head i was thinking of Test tools / Suite tools headings but inside also a smaller headings for type of action

src/swagger/client/functional-testing-types.ts
```diff
@@ -37,6 +37,91 @@ export const CancelFunctionalTestingSuiteExecutionSchema = z.object({
...
+      .array(RunApiTestsBlockSchema)
+      .min(1)
+      .describe(
+        'Required — ordered groups ("blocks") of tests to run one after another. ' +
+          "Must include at least one entry; suites cannot be created without a workflow. " +
+          "Within a block, tests run sequentially unless `parallel` is set. " +
+          "Block `title`s must be unique within the suite.",
+      ),
+  })
+  .superRefine((data, ctx) => {
```
KacperSB, line 93: Did you tested if that schema is enforced?
From what i checked `superRefine` produces `ZodEffects`, but the method responsible from registering schema: `schemaToRawShape` (in `src/common/server.ts`) only handles `ZodObject` and `ZodIntersection`. So for `ZodEffects` it will unwrap to `undefined` - whole schema is missing then and LLMP don't know which fields to send at least in theory - maybe he could deduce them
Worth checking this

src/swagger/client/functional-testing-types.ts
```diff
@@ -37,6 +37,91 @@ export const CancelFunctionalTestingSuiteExecutionSchema = z.object({
...
+  typeof CreateFunctionalTestingSuiteParamsSchema
+>;
+
+export const CreateFunctionalTestingSuiteResponseSchema = z.object({
+  id: z.number().describe("ID of the newly created suite"),
+  slug: z.string().describe("Slug of the newly created suite"),
+  url: z
+    .string()
+    .describe("Link to the created suite in Swagger Functional Testing UI")
+    .optional(),
```
KacperSB: Why this is optional? From what i see on aqa side, that `url` is required in response 
(https://github.com/runreflect/aqa/blob/5c341188214605322b17b4f7802aee3d8f2bff94/backend/clients/management/src/main/scala/run/reflect/aqa/management/model/v1/SuitesRestTypes.scala#L127)

KacperSB (commented):
@bartoszkosicki13 
Please verify that `superRefine` usage and optional on `url` param
I will check again later

Asmodeios (approved):

src/swagger/client/functional-testing-types.ts
bartoszkosicki13: Checked this — good catch to flag, but it's not an issue here. schemaToRawShape does only handle ZodObject/ZodIntersection, but the ZodEffects-wrapping behavior for .superRefine() was a Zod v3 thing —
  we're on Zod v4 (^4.3.6), where the architecture changed: refinements are now stored as "checks" directly on the original type, so CreateFunctionalTestingSuiteParamsSchema instanceof ZodObject stays 
  true and .shape is still accessible normally.

  Verified this:
  CreateFunctionalTestingSuiteParamsSchema instanceof ZodObject // true
  CreateFunctionalTestingSuiteParamsSchema.shape // { name, agentName, runApiTests }

  And that the superRefine validation itself (duplicate title check) still fires correctly on safeParse, returning the expected "Duplicate block title..." error.

  So the schema is extracted correctly and the LLM sees all fields in the JSON Schema.

src/swagger/client/functional-testing-types.ts
bartoszkosicki13: Confirmed BE always returns `url`, so I updated it — fixed, thanks!

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
bartoszkosicki13: Fair point — broke it up into a short intro sentence plus a bullet list for the block-level params, should be easier to scan now.

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
bartoszkosicki13: Agreed, that grouping (Test tools / Suite tools with action-type sub-headings) makes sense — let's tackle it as a follow-up PR rather than scope-creeping this one.
I updated ticket https://smartbear.atlassian.net/browse/RF-4941 to improve descriptions

docs/products/SmartBear MCP Server/swagger-functional-testing-integration.md
KacperSB: thx!

src/swagger/client/functional-testing-types.ts
KacperSB: Nice, you right, i checked wrong version

KacperSB (approved):
lgtm

PR #633, opened 2026-08-04
chore: edit workflow to remove versions with remote URLs from registry

szymon-pleszkun-sb:
## Goal                                                                                                                                                                                                    
                                                                                                                                                                                                             
  We need to free the remote URL `https://swagger.mcp.smartbear.com/mcp` so it                                                                                                                               
  can be published under a new `com.smartbear/swagger-mcp` registry entry —                                                                                                                                  
  it's currently still referenced by five older `com.smartbear/smartbear-mcp`                                                                                                                                
  versions (0.25.0, 0.25.1, 0.26.0, 0.26.1, 0.27.0), and the registry's                                                                                                                                      
  `validateNoDuplicateRemoteURLs` check rejects publishing a URL that's already                                                                                                                              
  claimed by any non-deleted version of another server. The MCP registry                                                                                                                                     
  maintainers declined to edit our entries themselves, so this needs a                                                                                                                                       
  self-serve fix.                                                                                                                                                                                            
                                                                                                                                                                                                             
  ## Design                                                                                                                                                                                                  
                                                                                                                                                                                                             
  The obvious approach — PUT an edited server body with the remote stripped                                                                                                                                  
  out — doesn't work for us: that endpoint requires `edit` permission, and                                                                                                                                   
  DNS/GitHub publisher auth (what we authenticate with) only ever grants                                                                                                                                     
  `publish`. Confirmed both in the registry source (only an admin-configured                                                                                                                                 
  OIDC provider or local dev auth grant `edit`) and empirically (a real                                                                                                                                      
  `--apply` run returned 403 on all 5 versions).                                                                                                                                                             
                                                                                                                                                                                                             
  The status-only `PATCH .../versions/{version}/status` endpoint is different:                                                                                                                               
  it accepts `publish` OR `edit` permission. Setting a version's status to                                                                                                                                   
  `deleted` removes it from the duplicate-remote-URL conflict scan (which                                                                                                                                    
  excludes only deleted versions), freeing the URL — with the same credentials                                                                                                                               
  we already use to publish, no admin involvement required.                                                                                                                                                  
                                                                                                                                                                                                             
  Built as a generic, parameterized CLI + `workflow_dispatch` wrapper (not a                                                                                                                                 
  one-off script) so the same tool covers future remote-URL migrations: target                                                                                                                               
  versions either by `--remote-url` (auto-detects every version that still                                                                                                                                   
  holds it) or an explicit `--versions` list, with a configurable `--status`                                                                                                                                 
  (default `deleted`) and `--message`. Defaults to a dry run; writes require                                                                                                                                 
  `--apply`.

  Two safety choices worth calling out:                                                                                                                                                                      
  - The registry also exposes a *bulk* `PATCH /servers/{name}/status` that sets                                                                                                                              
    **every** version's status in one call — including whatever is currently                                                                                                                                 
    `latest`. This tool deliberately never calls it; it only PATCHes specific                                                                                                                                
    versions one at a time.                                                                                                                                                                                  
  - It hard-refuses to touch any version flagged `isLatest`, however that                                                                                                                                    
    version was selected (explicit list or URL auto-detection), so a typo'd                                                                                                                                  
    `--versions` input can't take down the live published server.                                                                                                                                            
                                                                                                                                                                                                             
  ## Changeset                                                                                                                                                                                               
                                                                                                                                                                                                             
  - `scripts/registry-remove-remote.js` — CLI that changes the lifecycle status                                                                                                                              
    of specific registry server versions to free a remote URL. Auto-detects (or                                                                                                                              
    takes an explicit list of) affected versions, dry-run by default,                                                                                                                                        
    `--apply`/`--yes` to write, refuses to touch the current `isLatest` version,                                                                                                                             
    token resolved via `mcp-publisher`'s saved login.                                                                                                                                                        
  - `.github/workflows/registry-remove-remote.yaml` — manual `workflow_dispatch`                                                                                                                             
    wrapper with `server` / `remote_url` / `versions` / `status` / `message` /                                                                                                                               
    `apply` inputs. Inputs are passed through `env:` (never interpolated into                                                                                                                                
    the shell script) to avoid injection. `mcp-publisher` is pinned to `v1.7.9`                                                                                                                              
    rather than `latest`.                                                                                                                                                                                    
                                                                                                                                                                                                             
  ## Testing                                                                                                                                                                                                 
                                                                                                                                                                                                             
  - Dry run against the live registry correctly lists the 5 affected versions                                                                                                                                
    and the intended status transition for each.                                                                                                                                                             
  - Verified the safety guard: passing the current latest version (`0.33.0`)                                                                                                                                 
    alongside a real target in `--versions` aborts with a clear error instead                                                                                                                                
    of proceeding.                                                                                                                                                                                           
  - Traced the permission check in the registry source (`status.go`) confirming                                                                                                                              
    the status endpoint accepts `publish` permission, unlike the full edit                                                                                                                                   
    endpoint.                                                                                                                                                                                                
  - `biome check` clean; `--help` output verified.


scripts/registry-remove-remote.js
```diff
@@ -256,57 +286,59 @@ async function main() {
...
-  );
-
-  // Restrict to an explicit version list if provided.
-  if (args.versions) {
+  let candidates;
+  if (args.remoteUrl) {
+    candidates = allVersions.filter((rec) =>
+      serverHasRemote(rec.server, args.remoteUrl),
+    );
+  } else {
```
KacperSB: If someone passes `--remote-url X --versions a,b,c` then the `--versions` part is silently dropped.
I'm not in the context so maybe the drop is expected - if so, maybe we can just warn about that?

.github/workflows/registry-remove-remote.yaml
```diff
@@ -1,14 +1,19 @@
 name: Registry — Remove Remote URL
```
KacperSB: Maybe we can rename these scripts filename and inner name as they no longer remove a remote URL if i understand correctly

KacperSB (approved):

KrzysztofNocen (approved):

PR #639, opened 2026-08-05
sdkv1 to sdkv2 changes applied to bearq

dev-smartbear:
Applied codemod(sdkv2) modelcontextprotocol changes to src/bearq.

### DIFF
--- a/src/bearq/tool/tasks/wait-for-task.ts
+++ b/src/bearq/tool/tasks/wait-for-task.ts
@@ -1,7 +1,5 @@
-import type { ToolCallback } from "@modelcontextprotocol/sdk/server/mcp.js";
-import type { ZodRawShape } from "zod";
 import { z } from "zod";
-import { Tool, ToolError } from "../../../common/tools";
+import { Tool, ToolError, type ToolHandler } from "../../../common/tools";
 import type { ToolParams } from "../../../common/types";
 import type { BearQClient } from "../../client";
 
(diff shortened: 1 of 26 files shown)


src/bearq/tool/tasks/wait-for-task.ts
```diff
@@ -43,11 +41,11 @@ export class WaitForTask extends Tool<BearQClient> {
     if (!res.body)
       throw new ToolError(`GET /tasks/${taskId}/stream: no response body`);
 
-    const progressToken = extra?._meta?.progressToken;
+    const progressToken = ctx?.mcpReq._meta?.progressToken;
```
amoli13: [nit] We add optional chain in `mcpReq` also. Tests won't catch this as I see you pass `{ mcpRec: {} }`. Won't expect `mcpReq` to be absent but good for consistency.

amoli13 (approved):
Looks good to merge. I see you narrowed down the scope of this branch to just BearQ related changes than the diff shared on slack earlier. So just one thing, in the ticket I read "Node engines ≥20" should be mentioned in package.json so good to add that engine requirement.

src/bearq/tool/tasks/wait-for-task.ts
dev-smartbear: done

s-leirbag (approved):
smoke tested all tools ✅

PR #647, opened 2026-08-06
sdkv1 to sdkv2 changes applied to reflect

dev-smartbear:
Applied codemod(sdkv2) modelcontextprotocol changes to src/reflect.

### DIFF
--- a/src/reflect/tool/recording/connect-to-session.ts
+++ b/src/reflect/tool/recording/connect-to-session.ts
@@ -1,8 +1,6 @@
 import { randomUUID } from "node:crypto";
-import type { ToolCallback } from "@modelcontextprotocol/sdk/server/mcp.js";
-import type { ZodRawShape } from "zod";
 import { z } from "zod";
-import { Tool, ToolError } from "../../../common/tools";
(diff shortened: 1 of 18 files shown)


src/reflect/tool/recording/connect-to-session.ts
```diff
@@ -80,7 +78,7 @@ export class ConnectToSession extends Tool<ReflectClient> {
 
     // This identifies the MCP session, rather than the Reflect recording session.
     // There can be multiple MCP sessions if we're running in HTTP mode.
-    const mcpSessionId = extra?.sessionId;
+    const mcpSessionId = ctx?.sessionId;
```
mahavir1166: [nit] The new `ToolHandler` type makes `ctx` a required param, so the `?.` in `ctx?.sessionId` is no longer needed, only `sessionId` itself is optional.

mahavir1166 (approved):
🚢

src/reflect/tool/recording/connect-to-session.ts
dev-smartbear: done

PR #651, opened 2026-08-06
feat(functional-testing): Improve error messaging in failed requests

sergioifg94:
## Goal

Fixes https://smartbear.atlassian.net/browse/RF-5806

The MCP local server delegates on the Swagger Functional Testing public API to perform all operations. These requests are made using the `fetch` native function.

Currently, a `catch` clause throws a generic error if an exception is thrown by `fetch`. This is not ideal as it does not give the user enough information to troubleshoot the error. Possible causes for the `fetch` function to throw are networking or TLS issues, which can be the result of the user environment.

## Design

Attempt to extract the code from the error and display it in the error message. This allows the user and (depending on the model) the LLM to troubleshoot possible issues on their environment

## Changeset

* Error handling on the Swagger Functional Testing toolset

## Testing

* Created unit test with mocked server
* Locally, by purposefully misconfiguring the base URL and verifying that the error code is displayed. Below an example of the error message displayed now:
image


SebastianRomejko (approved):
It's fine as it is now (because any error handling is better than none), but I wonder if we couldn't improve it even more (for a user and/or LLM) by having specific messages for specific error codes. To be honest I'm not sure where to find exactly what errors might be thrown here, but in this page couple exceptions are mentioned with reasons: https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#exceptions
And here are common system errors: https://nodejs.org/api/errors.html#common-system-errors
Maybe we can use it to construct more specific messages.

But on the other hand maybe it's not so important and we can just have this message, which contains the code, so both LLM and user can verify exactly what it means.

sergioifg94:
Thank you @SebastianRomejko. What you're suggesting was actually my initial approach, a switch case with the most common possible codes:
```ts
switch (code) {
      // DNS / routing
      case 'ENOTFOUND':
      case 'EAI_AGAIN':      
        throw new ToolError(`DNS lookup failed: ${url}`);

      // TCP
      case 'ECONNREFUSED':
        throw new ToolError('Connection refused');
      case 'ECONNRESET':
        throw new ToolError('Connection reset by server');
      case 'ETIMEDOUT':
        throw new ToolError('Connection timed out');

      // TLS / certificate
      case 'CERT_HAS_EXPIRED':
      case 'CERT_UNTRUSTED':
      case 'DEPTH_ZERO_SELF_SIGNED_CERT':
      case 'SELF_SIGNED_CERT_IN_CHAIN':
      case 'UNABLE_TO_VERIFY_LEAF_SIGNATURE':
      case 'ERR_TLS_CERT_ALTNAME_INVALID':
        throw new ToolError(`TLS/cert error (${code})`);

      default:
        throw err; 
}
```

I eventually decided to keep it simple and just show the code in the error message, I noticed that the LLM is able to translate it to a user friendly message anyway.

I don't mind revisiting this approach if you think it's worth it though

src/swagger/client/functional-testing-api.ts
```diff
@@ -60,7 +60,16 @@ export class FunctionalTestingAPI {
...
     try {
       response = await fetch(`${this.baseUrl}/${relativePath}`, init);
-    } catch {
+    } catch (err: any) {
+      // An exception being thrown by fetch() can happen for networking or TLS issues. Attempt to
+      // extract the code to include it in the error message.
+      const code = err.cause?.code;
+      if (code) {
+        throw new ToolError(
+          `Failed to reach Swagger Functional Testing API: ${code}. Please verify your settings and network connectivity`,
```
KacperSB:
```suggestion
          `Failed to reach Swagger Functional Testing API: ${code}. Please verify your settings and network connectivity.`,
```

KacperSB (approved):

bartoszkosicki13 (approved):

bartoszkosicki13 (approved):

Issue #631, opened 2026-08-04
qtm4j_search_test_cases: filter.folders is ignored; schema inconsistent with other QTM4J tools (folderId)


chandrahaas-komprise:
## Summary

`qtm4j_search_test_cases` documents `filter.folders` (array of numeric folder IDs) for folder-scoped search, but the filter appears to have **no effect** on results. The response `total` and first page are identical to an unfiltered project-wide search.

Other QTM4J MCP tools (e.g. `qtm4j_search_test_cycles`, link/unlink tools) use **`folderId`** (singular number), not `folders`. This inconsistency may indicate the wrong filter field name is exposed for test case search.

## Environment

- Package: `@smartbear/mcp` **0.33.0** (via `npx -y @smartbear/mcp`)
- Client: Cursor MCP (`smartbear` server)
- QTM4J project: **KOMTEST** (`projectId` **10100**, Jira: `komprise.atlassian.net`)
- Folder under test: **Data Copy** (`folderId` **53554**)

## Steps to reproduce

1. Call `qtm4j_set_project_context` with `projectKey: "KOMTEST"`.
2. Call `qtm4j_search_test_cases` with:

```json
{
  "filter": {
    "projectId": "10100",
    "folders": [53554]
  },
  "fields": ["key", "summary", "status", "folders"],
  "maxResults": 50,
  "startAt": 0,
  "sort": "key:asc"
}
```

3. Compare with a baseline call that omits `folders` (only `projectId` / default context).

## Actual behavior

| Filter | `total` | First results |
|--------|---------|---------------|
| `{ "projectId": "10100" }` | **21359** | `KOMTEST-TC-8`, `TC-10`, … (`folders: null`) |
| `{ "projectId": "10100", "folders": [53554] }` | **21359** | **Same page** — no folder narrowing |
| `{ "projectId": "10100", "folders": [53554], "status": ["Automatable"] }` | **2173** | Status filter works; results span many folders (not only 53554) |

Known test case in folder 53554 (`KOMTEST-TC-2842`, summary: "Files copy failure - lock the file etc") is **not** returned on page 1 when `folders: [53554]` is set, even though its `folders` field in unfiltered paginated results is `[{ "id": 53554, "name": "Data Copy" }]`.

Passing `filter.folderId: 53554` is **silently dropped** by the Zod schema in `src/qtm4j/schema/get-test-case.schema.ts` (field not defined), so it cannot be used as a workaround via MCP.

## Expected behavior

[...]

PR #655, opened 2026-08-09
fix(qtm4j): fix folderId handling in create and search test case tools

### LINKED ISSUE #643: qtm4j_create_test_case: folderId silently ignored — new test cases always land in "MCP Generated"
## Summary

`qtm4j_create_test_case` accepts a `folderId` parameter (per its tool schema) to place a new test case in a specific folder, but the parameter has no effect. Every test case created via this tool lands in the project's default "MCP Generated" folder, regardless of what `folderId` is passed.

This looks related to #631, which reports a similar folder-handling inconsistency on `qtm4j_search_test_cases` (`filter.folders` vs `filter.folderId`). This issue is about the create path specifically.

## Environment

- Package: `@smartbear/mcp` (via Claude's SmartBear MCP connector)
- QTM4J project: KW, projectId 11200
- Target folder: test case KW-TC-96's folder (folder 2628513)
- Default folder new cases land in instead: "MCP Generated" (folder 2637830)

## Steps to reproduce

1. Call `qtm4j_set_project_context` with projectKey: "KW".
2. Look up an existing test case's folder, e.g. `qtm4j_search_test_cases` with filter.searchText: "KW-TC-96" and fields: ["key","summary","folders"] -> folder id 2628513.
3. Call `qtm4j_create_test_case`:

{
"folderId": 2628513,
"summary": "test case summary"
}

4. Search for the newly created key (e.g. KW-TC-98) and inspect its folders field.

## Actual behavior

The new test case (KW-TC-98) is created successfully, but its folders field comes back as:

"folders": [{ "name": "MCP Generated", "id": 2637830 }]

not the requested folder 2628513.

The tool's success response also gives no indication anything was overridden:

{
"id": "ZzoQc825uWpK1o",
"key": "KW-TC-98",
"versionNo": 1,
"summary": "test case summary"
}

No warning field, no folder confirmation, so the misplacement is silent and only discoverable by a follow-up search.

## Expected behavior

- Passing folderId to `qtm4j_create_test_case` should place the new test case in that folder.
- If the field is unsupported or mismatched against the underlying API, the tool should return a warning (consistent with how it already warns for unresolvable priority/status/labels/components) rather than silently defaulting to "MCP Generated".

## Suspected cause

Possibly the same root cause as #631, a mismatch between the MCP tool's exposed schema field name and what the underlying QTM4J create endpoint actually expects/maps to for folder placement.

## Workaround

None currently via MCP - test cases must be moved into the correct folder manually in the QMetry UI after creation.

Related: #631

varshini03:
## Goal

Fix `folderId` handling in `qtm4j_create_test_case` and `qtm4j_search_test_cases` so test cases are created in the right folder and search results are filtered correctly. Aimed to help fix #643, #631.

## Design

**Create:** `handle()` hardcoded `folderId: "MCP Generated"`, and `folderId` was in `FIELD_CONFIG` causing `CommonAttributeResolver` to treat the numeric ID as a name, fail the lookup, and delete it. Fixed with `parsed.folderId ?? "MCP Generated"` and removed `folderId` from `FIELD_CONFIG`.

**Search:** `SearchTestCaseFilter` exposed `folders: number[]` but the API expects `folderId: number`. Wrong field name was silently ignored, returning all test cases. Renamed to match every other QTM4J filter schema in this codebase.

## Changeset

| File | Change |
|---|---|
| `create-test-case.ts` | `folderId ?? "MCP Generated"`; removed `FOLDER` from `FIELD_CONFIG` |
| `get-test-case.schema.ts` | `folders: number[]` → `folderId: number` |
| `get-test-cases.ts` | Updated example to use `folderId` |
| `create-test-case.test.ts` | 3 tests |
| `get-test-cases.test.ts` | 1 test |
| `folder-handling.integration.test.ts` | New - integration tests for all bugs |

## Testing

TDD: failing tests written first to reproduce each bug, then fixed. Integration test uses a simulated resolver that deletes unresolvable fields (matching real `CommonAttributeResolver` behaviour) to prove `folderId` survives the full pipeline. Full test suite passing.

## Notes to maintainers/codeowners

- `FOLDER_IDS` in `constants.ts` is defined but unused - multi-folder filtering may have been intended. If the API supports `folderIds: number[]`, that might need a separate follow-up. The integration test proves the correct field reaches the API; filtering behaviour was verified manually via MCP Inspector. 
- **Open to feedback/clarifiication on understanding/expectations; happy to make any changes needed.**
- All other QTM4J schemas use `folderId: number` (singular), suggesting the API supports one folder at a time. Multi-folder filtering would need API contract verification first, left as a separate follow-up.
- No access to a live QTM4J account; validated through code analysis, TDD, and integration tests.

### DIFF
--- a/src/common/__snapshots__/server-definitions.test.ts.snap
+++ b/src/common/__snapshots__/server-definitions.test.ts.snap
@@ -10032,17 +10032,14 @@ Expected Output: Test cases with all available fields explicitly requested
 \`\`\`json
 {
   "filter": {
-    "folders": [
-      123,
-      456
-    ],
+    "folderId": 123,
     "fixVersions": [
       789
     ]
   }
 }
 \`\`\`
-Expected Output: Test cases in the specified folders and fix versions
+Expected Output: Test cases in the specified folder and fix versions
 
 11. Complex filter: multiple criteria combined with multi-field sort
 \`\`\`json
--- a/src/qtm4j/schema/get-test-case.schema.ts
+++ b/src/qtm4j/schema/get-test-case.schema.ts
@@ -43,12 +43,12 @@ export const SearchTestCaseFilter = zod
         "Component names to include (OR logic within array). Example: ['UI', 'Cloud', 'API']. " +
           "Use exact component names as configured in the project.",
       ),
-    folders: zod
-      .array(zod.number())
+    folderId: zod
+      .number()
       .optional()
       .describe(
-        "Folder IDs (numeric, OR logic within array). Example: [123, 456]. " +
-          "Retrieve folder IDs from the project's folder structure.",
+        "Folder ID to filter test cases by. Example: 123. " +
+          "Retrieve the folder ID by right-clicking the target folder in QTM4J and selecting 'Copy Folder Id'.",
       ),
     assignee: zod
       .array(zod.string())
--- a/src/qtm4j/tool/test-case/create-test-case.test.ts
+++ b/src/qtm4j/tool/test-case/create-test-case.test.ts
@@ -69,9 +69,9 @@ describe("CreateTestCase", () => {
 
       const result = await instance.handle(rawArgs);
 
-      // FIELD_CONFIG has 4 entries: PRIORITY, STATUS, COMPONENTS, LABELS, FOLDER
-      expect(mockFieldResolver.getResolver).toHaveBeenCalledTimes(5);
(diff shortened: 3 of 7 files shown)


mes-dfr:
@varshini03 why was this closed? Has the bug been fixed in a different PR/release?

varshini03:
This got auto-closed when I accidentally deleted my fork, not intentional, and the bug isn't fixed elsewhere. let me reraise

PR #660, opened 2026-08-12
feat: Add Swagger patch api tool

lukasikp:
## Goal

Add a `swagger_patch_api` tool that applies targeted edits to an existing API definition in SwaggerHub Registry. This lets agents fix specific issues without regenerating the whole definition, which is faster and avoids unrelated changes.

## Design

- Edits are exact search/replace pairs on the YAML text, so the rest of the definition stays byte-for-byte untouched.
- The operation is atomic: nothing is saved unless every edit applies. Failed edits are returned with `no_match` or `ambiguous` status so the agent can retry.
- Optional `newVersion` saves the result as a new version. It is rejected if the version already exists, and `info.version` is updated automatically.

## Changeset

- New `swagger_patch_api` tool with input/output schemas and `WRITE_DESTRUCTIVE` annotations.
- New `patch-spec-utils.ts` with `applyEdits` and `setVersionToYamlSpec` helpers.
- `swagger_get_api_definition` gets a `format` param (`json`/`text`); `text` returns raw YAML needed as the source for patch edits.

## Testing

- Unit tests for `applyEdits` and `setVersionToYamlSpec` (unique match, ambiguous, no match, replaceAll, quoting styles).
- Unit tests for `patchApi` covering the happy path, failed edits, and existing `newVersion` rejection.
- Updated tool definition snapshots.


kacperdufrat (approved):

VarshaNagtilak1 (approved):

src/swagger/client/api.ts
```diff
@@ -1521,4 +1534,72 @@ export class SwaggerAPI {
...
+      const exists = await this.apiVersionExists({
+        owner: params.owner,
+        apiName: params.apiName,
+        version: params.newVersion,
+      });
+      if (exists) {
+        throw new ToolError(
+          `Version ${params.newVersion} already exists for ${params.owner}/${params.apiName}. Choose a different version name or omit newVersion to patch in place.`,
+        );
+      }
```
KacperSB, line 1594: There is a short window between this check for version existence and moment of using `createOrUpdateAPi`. 
At the point of the check the version could not exists and the code will move on but just after it, someone could create new version within that window, resulting in no check at all.

Just something to be aware of as i don't see immediate solution for it, could be an edge case

src/swagger/client/api.ts
```diff
@@ -1521,4 +1534,72 @@ export class SwaggerAPI {
...
+    const { text: patchedText, failed } = applyEdits(text, params.edits);
+    text = patchedText;
+
+    if (failed.length > 0) {
+      return { failed, saved: false };
+    }
+
+    text = setVersionToYamlSpec(text, targetVersion);
+
+    const saveResult = await this.createOrUpdateApi({
```
KacperSB: I also see that `createOrUpdateApi` always marks the API private, due to `searchParams.append("isPrivate", "true");`.
So question is if i patch the public API, should it mark it as private? 
If that's expected, it could be at least documented in the tool itself to not create confusion IMO

src/swagger/client/registry-types.ts
```diff
@@ -314,3 +368,41 @@ export const SearchApisOutputSchema = z.object({
...
+  url: z.string().optional(),
+});
+
+// Response type
+export interface PatchApiFailedEdit {
+  index: number;
+  oldString: string;
+  error: "no_match" | "ambiguous";
+  matchCount: number;
+}
```
KacperSB: Will this ever change without change on the `PatchApiFailedEditSchema`?
Asking as it express the exact same shape, so currently both can drift. If that should not happen, you could simplify it to be: `export type PatchApiFailedEdit = z.infer<typeof PatchApiFailedEditSchema>;`

src/swagger/client/api.ts
```diff
@@ -1521,4 +1534,72 @@ export class SwaggerAPI {
 
     return updated;
   }
+
+  async apiVersionExists({
```
KacperSB: [nit] could be `private`

src/swagger/client/api.ts
lukasikp: good catch Kacper! I updated code. Patch doesn't change api visibility and we will modify `createOrUpdateApi` too

bartosz-nowacki (approved):

PR #661, opened 2026-08-12
refactor(functional-testing): tool annotations

SebastianRomejko:
## Goal

Functional Testing tools (`src/swagger/client/functional-testing-tools.ts`) only set readOnly + idempotent inline; destructive and openWorld were omitted and silently fell back to the registration defaults in src/common/server.ts (destructiveHint: false, openWorldHint: false) regardless of what the tool actually does. MCP clients (Claude Desktop, Claude Code, etc.) rely on these annotation hints to gate confirmation prompts, sandboxing, and tool visibility, so a mutating/destructive tool that reports no destructive or open-world behavior is misleading and unsafe. This change brings the Functional Testing tools in line with the annotation pattern already used by the rest of the Swagger tools.

## Design

Reused the shared `READ_ONLY` / `WRITE` / `WRITE_DESTRUCTIVE` presets from `tool-constants.ts` instead of repeating inline flags, matching the established pattern in `tools.ts`. Each preset fixes `readOnly` / `destructive` / `openWorld` together, so a tool can't set one hint and silently default the rest. Hints were reviewed per tool rather than mechanically swapped:

- Read tools (List Tests, Get Test Status, List Suite Executions, List Suites, Get Suite Status, Get Test Execution History) → `READ_ONLY`
- Create tools (Create Test, Create Suite) → `WRITE`
- Run Test / Run Suite → `WRITE` plus openWorld: true, since they execute test steps against arbitrary user-defined external endpoints — genuinely open-world side effects beyond the SmartBear backend.
- Cancel Suite Execution → `WRITE_DESTRUCTIVE`, since it irreversibly stops an in-progress run and should trigger a client confirmation prompt.

`idempotent: false` is kept explicit on all mutating tools (create/run/cancel), because the presets don't carry idempotency and the server default is true — without it, non-idempotent mutations would be mislabeled.

Changeset

src/swagger/client/functional-testing-tools.ts:

- Import `READ_ONLY`, `WRITE`, `WRITE_DESTRUCTIVE` from `./tool-constants`.
- Replace inline readOnly / idempotent flags on all 11 Functional Testing tools with the appropriate shared preset spread.
- Add openWorld: true to Run Test and Run Suite.
- Mark Cancel Suite Execution as destructive via WRITE_DESTRUCTIVE.

Testing
- tsc --noEmit passes with no errors
- smoke tested SFT tools

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- [Swagger] Corrected tool annotation hints for the Functional Testing tools so they no longer fall back to registration defaults: `run_test` and `run_suite` are now declared open-world (`openWorldHint: true`), `cancel_suite_execution` is declared destructive (`destructiveHint: true`), and all tools use the shared `READ_ONLY`/`WRITE`/`WRITE_DESTRUCTIVE` presets to set `readOnlyHint`/`destructiveHint`/`openWorldHint` explicitly.
+
 ## [0.35.0] - 2026-08-11
 
 ### Added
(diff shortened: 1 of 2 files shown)


KacperSB (approved):
amazing

Kamil-Gumieniuk (approved):
nice

bartoszkosicki13 (approved):

PR #677, opened 2026-08-20
cache issue fixed

SB-rohandhiman:
## Goal

Prevent cross-account data leakage in Bugsnag caching by removing per-session cache reliance and ensuring cache entries are isolated by caller identity.

## Design

Use a single process-wide cache for the Bugsnag client, but namespace every cache key with a hash of the authenticated caller’s token. This keeps cache reuse efficient while preventing one account from reading another account’s cached org, project, or field data. The implementation also prefers the current request’s auth header when present, so HTTP requests resolve the correct namespace per call.

## Changeset

- Replaced `server.getCache()` usage with a shared module-level `CacheService`.
- Added lazy initialization for the shared cache instance.
- Added `getProjectApiKey()`, which resolves the Bugsnag project API key from the request header first, then falls back to the configured value.
- Added credential-based cache namespacing using a SHA-256 hash of the auth token.
- Updated cache reads/writes in:
  - `getOrganization()`
  - `getProjects()`
  - `getCurrentProject()`
  - `getProjectEventFields()`
  - `getProjectTraceFields()`
  - project lookup/write paths
- Added safeguards so unauthenticated callers bypass the shared cache entirely.
- Expanded tests to cover:
  - shared cache behavior
  - cache key namespacing per account
  - request-header-driven namespace resolution
  - no-cache behavior without credentials
  - prevention of cross-account key collision

## Testing

- I have tested with : unit testcases
- Verified cache keys are namespaced instead of using raw logical keys.
- Added isolation tests for multiple accounts and request-scoped auth headers.
- Added coverage for unauthenticated requests bypassing cache access.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- [Bugsnag]: Prevent cross-account data leakage in Bugsnag caching by namespacing cache keys with the caller's credential hash. [#677]
+
+### Changed
+
+- [Bugsnag] Use a single process-wide `CacheService` and namespace every cache key with a SHA-256 hash of the authenticated caller's token.
+- [Bugsnag] Prefer the request auth header when resolving cache namespace via `getProjectApiKey()` so per-request auth is honored.
+- [Bugsnag] Ensure unauthenticated callers bypass the shared cache to avoid exposing cached data.
+- [Bugsnag] Updated cache usage in `getOrganization()`, `getProjects()`, `getCurrentProject()`, `getProjectEventFields()`, and `getProjectTraceFields()`.
+
 ## [0.37.0] - 2026-08-20
 - [QMetry]: Added Test Case, Test Suite, Issue module, Create/Update UDF capability Added. [#666](https://github.com/SmartBear/smartbear-mcp/pull/666)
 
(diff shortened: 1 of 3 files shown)


dev-smartbear (commented):
with no project API key configured, CURRENT_PROJECT is keyed as tok:<hash>:bugsnag_current_project:- ,so all sessions on the same auth token now share one "current project" slot and overwrite each other's selection (previously per-session) not a leak, same account, but it breaks two concurrent sessions working on different projects
can this one key get a session discriminator, or is sharing it intended?

SB-rohandhiman:
> with no project API key configured, CURRENT_PROJECT is keyed as tok::bugsnag_current_project:- ,so all sessions on the same auth token now share one "current project" slot and overwrite each other's selection (previously per-session) not a leak, same account, but it breaks two concurrent sessions working on different projects can this one key get a session discriminator, or is sharing it intended?

Great catch! This is a valid concern we're addressing as an incremental enhancement.
Currently, we prioritized solving the memory leak issue. 
The limitation only affects concurrent sessions with the same token and no project_api_key — 
not a security leak, but we understand it breaks that specific workflow.
**Creating a ticket** to implement per-session discrimination for CURRENT_PROJECT in the next iteration.
Thanks for the review!

dev-smartbear (dismissed):

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- [Bugsnag]: Prevent cross-account data leakage in Bugsnag caching by namespacing cache keys with the caller's credential hash. [#677]
+
+### Changed
+
+- [Bugsnag] Use a single process-wide `CacheService` and namespace every cache key with a SHA-256 hash of the authenticated caller's token.
+- [Bugsnag] Prefer the request auth header when resolving cache namespace via `getProjectApiKey()` so per-request auth is honored.
```
dev-smartbear: shouldn't this be 

-- Prefer the request auth header when resolving cache namespace via `getAuthToken()` so per-request auth is honored -- ?

PR #679, opened 2026-08-21
cache issue fixed

dev-smartbear:
## Goal

Prevent cross-account data leakage in Bugsnag caching by removing per-session cache reliance and ensuring cache entries are isolated by caller identity.

## Design

Use a single process-wide cache for the Bugsnag client, but namespace every cache key with a hash of the authenticated caller’s token. This keeps cache reuse efficient while preventing one account from reading another account’s cached org, project, or field data. The implementation also prefers the current request’s auth header when present, so HTTP requests resolve the correct namespace per call.

## Changeset

- Replaced `server.getCache()` usage with a shared module-level `CacheService`.
- Added lazy initialization for the shared cache instance.
- Added `getProjectApiKey()`, which resolves the Bugsnag project API key from the request header first, then falls back to the configured value.
- Added credential-based cache namespacing using a SHA-256 hash of the auth token.
- Updated cache reads/writes in:
  - `getOrganization()`
  - `getProjects()`
  - `getCurrentProject()`
  - `getProjectEventFields()`
  - `getProjectTraceFields()`
  - project lookup/write paths
- Added safeguards so unauthenticated callers bypass the shared cache entirely.
- Expanded tests to cover:
  - shared cache behavior
  - cache key namespacing per account
  - request-header-driven namespace resolution
  - no-cache behavior without credentials
  - prevention of cross-account key collision

## Testing

- I have tested with : unit testcases
- Verified cache keys are namespaced instead of using raw logical keys.
- Added isolation tests for multiple accounts and request-scoped auth headers.
- Added coverage for unauthenticated requests bypassing cache access.


src/bugsnag/client.ts
```diff
@@ -106,18 +130,81 @@
...
+  // customer apps, so namespacing on it would let anyone who knows a project
+  // key read another account's cached org and project list.
+  // Returns null when there is no credential at all — callers then bypass the
+  // cache rather than sharing a single anonymous bucket.
+  private cacheNamespace(): string | null {
+    const authToken = this.getAuthToken();
+    if (!authToken) {
+      return null;
+    }
+    return `tok:${createHash("sha256").update(authToken).digest("hex").slice(0, 32)}`;
```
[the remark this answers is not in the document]
SB-rohandhiman: We’ve updated the security code and are testing it now. We’ll share an update once testing is complete.

src/bugsnag/client.ts
```diff
@@ -106,30 +127,105 @@
...
+  // Returns null when there is no credential at all — callers then bypass the
+  // cache rather than sharing a single anonymous bucket.
+  private cacheNamespace(): string | null {
+    const authToken = this.getAuthToken();
+    if (!authToken) {
+      return null;
+    }
+    const cacheAuthority = this.getRequestApiAuthority();
+    const namespace = createHmac("sha256", "bugsnag-cache-ns")
+      .update(cacheAuthority)
```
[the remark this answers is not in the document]
SB-rohandhiman: We have fixed this issue.

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- [Bugsnag]: Prevent cross-account data leakage in Bugsnag caching by namespacing cache keys with the caller's credential hash. [#677](https://github.com/SmartBear/smartbear-mcp/pull/677)
```
kevinrvaz: this should be 679

src/bugsnag/client.ts
```diff
@@ -106,30 +127,105 @@ export class BugsnagClient implements Client {
...
+  }
+
+  // Namespaced cache write. No-ops when there is no credential to namespace
+  // by, so an unauthenticated caller can never populate a shared entry.
+  private cacheSet<T>(key: string, value: T): void {
+    const namespacedKey = this.cacheKey(key);
+    if (!namespacedKey) {
+      return;
+    }
+    getSharedCache().set(namespacedKey, value);
```
kevinrvaz: we should also set a ttl, otherwise the data in the cache can go stale and always return same results, eg if we first fetch projects from bugsnag and then create a new project, the next list projects call will be missing the newly created project since we never invalidated the cache

kevinrvaz: also add some tests to check this behaviour

SB-rohandhiman: Hi @kevinrvaz We’re already using the existing CacheService, which applies a TTL to cached values. It defaults to 24 hours through CACHE_TTL, so the data won’t stay cached forever. However, the project list could still be stale for up to 24 hours.

CHANGELOG.md
SB-rohandhiman: Fixed. I updated the PR reference from #677 to #679.

src/bugsnag/client.ts
```diff
@@ -106,30 +127,105 @@ export class BugsnagClient implements Client {
...
+    const namespace = this.cacheNamespace();
+    if (!namespace) {
+      return null;
+    }
+    return `${namespace}:${key}`;
+  }
+
+  // Namespaced cache read. Returns undefined when there is no credential to
+  // namespace by, so the caller falls through to the API.
+  private cacheGet<T>(key: string): T | undefined {
```
kevinrvaz, line 161: make the return value consistent in some functions we are returning null and in some undefined

src/bugsnag/client.ts
kevinrvaz: isnt 24hrs too long? ull get wrong results for a day.

kevinrvaz: and the user has no way to bust the cache themselves

kevinrvaz: and the user has no way to bust the cache themselves

kevinrvaz: another option is to invalidate cache entries on create/update/delete operations, but if someone uses non-mcp operations then the cache will still be stale, a shorter ttl will still be better

SB-rohandhiman: Fixed. I added a 60-second TTL.

src/bugsnag/client.ts
SB-rohandhiman: Fixed. The internal cache helpers now consistently return undefined when no value is available, matching CacheService.get().

src/bugsnag/client.ts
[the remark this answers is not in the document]
SB-jigneshR: As it is fixed by @SB-rohandhiman , dismissing Alert.

SB-jigneshR (approved):

kevinrvaz (approved):

SB-jigneshR:
Merging it as all checks have passed and have required approval.

PR #683, opened 2026-08-24
fix(functional-testing): clarify slug/id usage for suite tools

ernest-bednarczyk:
## Goal

[RF-5804](https://smartbear.atlassian.net/browse/RF-5804): LLM agents were observed misusing Swagger Functional Testing suite identifiers in two ways:
1. Calling `create_suite` with a numeric value (e.g. a test ID) as the suite `name`, instead of a meaningful, human-readable name.
2. Passing the suite's numeric `id` (returned by `create_suite`/`list_suites`) as the identifier to `run_suite`, even though those endpoints only accept the suite's `slug`.

## Design

- For (1), the tool description was clarified so the calling agent knows to use the name explicitly provided by the user, or fall back to a descriptive name derived from the tests being grouped into the suite.
- For (2), the ambiguous `suiteId` parameter was renamed to `slug` across all suite-execution tools, and the redundant numeric `id` was dropped from `create_suite` and `list_suites` responses entirely, since `slug` is the only identifier ever needed. This removes the confusing dual-identifier surface so agents can no longer pick the wrong one.

## Changeset
- `src/swagger/client/functional-t`name` field description in`CreateFunctionalTestingSuiteParam` → `slug` on`ListFunctionalTestingSuiteExecutinalTestingSuiteExecutionSchema`,`RunFunctionalTestingSuiteParamsScngSuiteExecutionSchema`, and`ListSuiteExecutionsResponse`; drotionalTestingSuiteResponseSchema`and `Suite`.- `src/swagger/client/functional-tistSuiteExecutions`,`cancelSuiteExecution`, `runSuite`to use `slug` instead of `suiteId`in validation, URL construction, aSuite` now returns only `slug` and`url`.
- `src/swagger/client/functional-testing-tools.ts`: updated tool summaries to reference `slug` instead of `suiteId`, and to note that `create_suite`/`list_suites` return the `slug` used to identify a suite in other tools.                                                               - `docs/products/SmartBear MCP Serting-integration.md`: updated docsto reflect the `slug`-only identifom suite responses.- `src/swagger/functional-testing/t.ts`,`src/swagger/functional-testing/swupdated tests/mocks for the`suiteId` → `slug` rename and removal of `id`.
- `CHANGELOG.md`: added entries under `## Unreleased` / `### Changed` for both fixes.

## Testing

- Verified via manual tests in claude code, **Full suite flow tested engaging all suite related tools**
- `npm test` (unit tests updated for the `slug` rename)

[RF-5804]: https://smartbear.atlassian.net/browse/RF-5804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/swagger/client/functional-testing-types.ts
```diff
@@ -71,7 +71,15 @@ export const RunApiTestsBlockSchema = z.object({
 
 export const CreateFunctionalTestingSuiteParamsSchema = z
   .object({
-    name: z.string().describe("Name for the new suite").trim().min(1),
+    name: z
+      .string()
+      .describe(
+        "Name for the new suite. Use the name explicitly provided by the user, or if none was given, " +
+          "a descriptive name derived from the purpose or context of the tests being grouped into this suite. " +
+          "This must be a human-readable name, not a numeric ID or test ID.",
```
KacperSB, line 79: This is not what we need to fix in fact, although it's a nice addition that we could keep.

Creation was in general correct before - we pass a name for the Suite to be created, and it is created with it, returning response with id (numeric suite id we don't care about) and slug (the name we passed).

Image

Image

In the ticket the problem is not about creating the Suite but running it (Kamil tried to create a suite with a name and run it *immediately* after). 

The actual issue here is about tool that runs given Suite, which tends to use the `id` instead of `slug` and i believe we need to update the `run_suite` tool descriptions to fix it properly.

Image

It could be bc in `run_suite` tool we use `suiteId` arg name for the slug arg - maybe we should rename it or describe it a little different 🤔 not sure - you can experiment with it

KacperSB (changes requested):

src/swagger/client/functional-testing-types.ts
ernest-bednarczyk: Oh! Missunderstood the problem. Thanks for clarification, will handle it.

ernest-bednarczyk:
@KacperSB I addressed the issue and verified all suite tools based on that. Id/slug were confusing both for agents and the codebase itself. IMO there is no point to operate on numeric ids at all.

KacperSB (commented):
@ernest-bednarczyk 
No issues with the PR but i'm wondering if we should also do the same `return { slug: data.slug, url: data.url };` for the `list_suites` tool? As you could run a Suite not only after it was created but also after you listed them. Currently the `list_suites` tool still returns `id` which can be misused. I think that wouldn't be much more effort and would solve the whole issue

ernest-bednarczyk:
@KacperSB really good point. I shifted all tools to operate on slugs only.

CHANGELOG.md
```diff
@@ -17,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Bugsnag] Ensure unauthenticated callers bypass the shared cache to avoid exposing cached data.
 - [Bugsnag] Updated cache usage in `getOrganization()`, `getProjects()`, `getCurrentProject()`, `getProjectEventFields()`, and `getProjectTraceFields()`.
 
+
```
KacperSB:
```suggestion
```

src/swagger/client/functional-testing-types.ts
```diff
@@ -197,22 +212,22 @@ export interface ListTestsResponse {
 }
 
 export interface Suite {
-  id: string;
-  accountId: number;
   name: string;
   slug: string;
   created: number;
-  numTestInstances: number;
-  url: string;
```
KacperSB, line 206: Is this correct that we also drop `accountId`, `numTestInstances` and `url` ?

src/swagger/client/functional-testing-types.ts
```diff
@@ -197,22 +212,22 @@ export interface ListTestsResponse {
...
 }
 
 export interface ListSuitesResponse {
   suites: Suite[];
-  stats?: {
-    executions: number;
-    passRate: number;
-    avgRuntimeSecs: number;
-    cumExecTimeSecs: number;
```
KacperSB, line 215: Same question here

KacperSB (commented):
I tested all Suite tools, looks like the issues is fixed at all levels

Left few question/suggestions for the code
Also please run `npm run format` to fix the PR checks issues

src/swagger/client/functional-testing-types.ts
ernest-bednarczyk: Yes, intentional. Those fields were never in the API response to begin with. Confirmed from the backend: SuitesService.scala#L126 constructs ListSuitesResponse with only name, slug, and created per suite. [aqa](https://github.com/runreflect/aqa/blob/17c93c38fd650fee26b6da21273b03c44745d3a6/backend/management/src/main/scala/run/reflect/aqa/management/server/suites/SuitesService.scala#L128)

Confirmed on main.

src/swagger/client/functional-testing-types.ts
ernest-bednarczyk: same as above

src/swagger/client/functional-testing-types.ts
KacperSB: a, indeed, OK

KacperSB (approved):
Nice!

PR #684, opened 2026-08-24
feat:  update portal product tool with url (SWG-21435)

VarshaNagtilak1:
## Goal

Update update_portal_product with url

## Design

Once product updated it should return the metadata of updated product with product url so agent can use that url.
## Changeset

Extended with url

## Testing

Test update_portal_product  retruns the url as well.


bartosz-nowacki (approved):

kacperdufrat (approved):

lukasikp (approved):

src/swagger/client/portal-types.ts
```diff
@@ -717,6 +717,7 @@ export const ProductOutputSchema = z.looseObject({
   description: z.string().optional(),
   public: z.boolean().optional(),
   hidden: z.boolean().optional(),
+  url: z.string().optional(),
```
KacperSB, line 720: [optional]
New `url` is added here and I see that `ProductOutputSchema` is shared not only for update but also for create and get portal tools. In this PR however `url` is only populated for the update action properly. 
This is true to ticket scope, although lack of population could make LLM hallucinate `url` for create/get tools or suggest it for users.
IF that could be potential issue - you could move the `url` variant to separate output schema only for update tool (splitting schemas per tool would be the best).

OR - you could consider populating `url` for other tools. IMO it's weird that only update returns the `url`, and not create or get, but if that's the design - it's fine as is

src/swagger/client/api.ts
```diff
@@ -607,9 +607,32 @@ export class SwaggerAPI {
...
-      success: true,
-    } as SuccessResponse);
+    const result = await this.handleResponse<Product | SuccessResponse>(
+      response,
+      {
+        success: true,
+      } as SuccessResponse,
+    );
+
+    if (hasId(result) && "slug" in result && "portalId" in result) {
```
KacperSB, line 617: [optional]
```suggestion
    if (hasId(result) && "slug" in result && result.portalId) {
```
Looking at `Product` type, `portalId` is typed as `string` or `undefined` - which makes that `"portalId" in result` validation not enough as `undefined` will return `true` for that check. 
Later it will be passed as `this.getPortal("undefined")` leading to 404 caught, but you could verify it earlier.

src/swagger/client/api.ts
```diff
@@ -607,9 +607,32 @@ export class SwaggerAPI {
...
+        const portalDetails = await this.getPortal(String(product.portalId));
+        const url = buildPortalLiveUrl(
+          this.config,
+          portalDetails,
+          product.slug,
+          null,
+          null,
+          false,
+        );
+        return { ...product, url } as Product & { url: string };
```
KacperSB, line 629: Could be:
```suggestion
        return { ...product, url } as Product;
```
Should be enough since `Product` has `[key: string]: unknown` so any extra property is already allowed.
This also is not true to other changes since `url` is defined as `optional`, but here it's explicit

KacperSB (approved):

src/swagger/client/portal-types.ts
VarshaNagtilak1: yes, In my opinion url should be return for create product as well. I will be going to ask this to product once back from PTO.

KacperSB: sure, thx

PR #697, opened 2026-09-01
feat : [QTM8-23150]: Security Vulnerability with QMetry MCP Tools

darshakpatel9:
## Goal

Security hardening and tool schema cleanup for the QMetry MCP integration, following up on the OAuth + AI Agent tools added in #637.

Addresses:
- **QTM8-24380** — Critical: SSRF via `baseUrl` tool parameter leading to Bearer token theft
- **QTM8-24382** — OAuth authentication bypass via `Qmetry-Token` / `apikey` header fallbacks on HTTP transport
- **QTM8-24386** — Path traversal via `fileName` parameter in `exportHtmlReport`

## Design

**Auth hardening:** `getToken()` in #637 implemented a three-tier fallback: OAuth Bearer → `Qmetry-Token`/`apikey` headers → configured token. On HTTP transport, tier 2 allowed any non-empty string to establish a valid MCP session without OAuth. Since `Qmetry-Token`/`apikey` headers are unreachable on stdio (no HTTP request context) and unauthorized on HTTP, the entire fallback tier is removed. `getToken()` now resolves: OAuth Bearer → configured token only.

**Path traversal mitigation:** `fileName` in `ExportHtmlReportArgsSchema` was a bare `z.string()`. Now enforces `.max(255)`, `.regex(/^[a-zA-Z0-9._-]+$/)` (alphanumeric + `.` `_` `-` only, no path separators), and `.refine()` against `..` sequences.

**SSRF / `baseUrl` removal:** Every QMetry tool schema accepted a per-request `baseUrl: z.string().url()` with no allowlist. A caller could set `baseUrl` to an attacker-controlled host, and the server would issue the outbound request there — including the `Authorization: Bearer <token>` header — leaking a live JWT. This was exploited end-to-end (SSRF callback captured a valid Bearer token, replayed successfully against the real QMetry backend). Fix: `baseUrl` is removed entirely from every QMetry tool schema. `resolveContext()` now always uses `this.endpoint` (resolved once from client configuration at startup), so there is no per-request path to redirect outbound requests or the token they carry. Confirmed this was QMetry-specific — the other 8 registered clients (Zephyr, Bugsnag, Swagger, Reflect, PactFlow, BearQ, QTM4J, Collaborator) only expose `baseUrl`/`base_url` as server-level `ConfigurationSchema` fields, never as a tool-call argument.

### Security — SSRF / Bearer Token Theft (QTM8-24380, Critical)
- `src/qmetry/types/common.ts` — Removed `baseUrl` field from `CommonFields` and from all 30+ tool schemas that referenced it.
- `src/qmetry/types/ai-agent.ts` — Removed `baseUrl` from the 3 AI Agent tool schemas.
- `src/qmetry/types/udf.ts` — Removed `baseUrl` from UDF tool schemas.
- `src/qmetry/client.ts` — `resolveContext()` now returns `baseUrl: this.endpoint` unconditionally instead of `args.baseUrl ?? this.endpoint`, eliminating the per-request override path entirely.

### Security — Auth (QTM8-24382)
- `src/qmetry/client.ts` — Removed `Qmetry-Token` and `apikey` header fallbacks from `getToken()`. Only OAuth Bearer and configured token remain.
- `src/qmetry/client.test.ts` — Updated tests to verify `Qmetry-Token`/`apikey` headers are now ignored (fall through to configured token).

### Security — Path Traversal (QTM8-24386)
- `src/qmetry/types/ai-agent.ts` — Added `.max(255)`, `.regex(/^[a-zA-Z0-9._-]+$/)`, and `..` refine guard on `fileName` in `ExportHtmlReportArgsSchema`.
- `src/qmetry/api/ai-agent.test.ts` — Added validation tests for `fileName`: accepts valid names, rejects traversal (`../`), path separators, spaces, shell metacharacters, empty strings, and >255 char names.

## Testing

All 2713 tests pass. Two existing `getToken` tests updated to reflect the removed fallback behavior.

```
Test Files  168 passed (168)
     Tests  2713 passed (2713)
```


darshakpatel9:
Screenshot 2026-09-01 at 18 11 45

varis-khan (approved):
Approved

PR #704, opened 2026-09-04
fix(functional-testing): update and fix tool descriptions

KacperSB:
## Goal

This PR updates and fixes Swagger Functional Testing tool descriptions, both internal and external so that:
- both follow good practices for various LLMs
- both reflect proper uses cases for each tool
- external are properly split and readable for both humans and LLMs

## Testing

- [x] verified regression for other swagger tools
- [x] verified regression for each SFT tool

### DIFF
--- a/src/swagger/client/functional-testing-tools.ts
+++ b/src/swagger/client/functional-testing-tools.ts
@@ -21,6 +21,7 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
     summary:
       "Lists all API tests available in your Swagger Functional Testing account. " +
       "Use this tool when you need to discover available tests before running them or checking their status. " +
+      "Returns test IDs usable with `swagger_run_test` to execute a single test, or passable to `swagger_create_suite` to group tests into a suite. " +
       "Do not use this tool to retrieve test execution results or history.",
     handler: "listFunctionalTestingTests",
     idempotent: true,
@@ -65,8 +66,8 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
     toolset: "Functional Testing",
     summary:
       "Get the status of a Swagger Functional Testing test execution. " +
-      "It returns information about the execution such as its status (running, passed or failed), run time, " +
-      "as well as the break down of the status of each test step.",
+      "Returns the execution status (`running`, `passed`, or `failed`), run time, and a per-step breakdown. " +
+      "Poll this tool after calling `swagger_run_test` until status is `passed` or `failed`; `running` means the execution is still in progress.",
     inputSchema: GetFunctionalTestingExecutionTestSchema,
     handler: "getFunctionalTestingExecution",
     idempotent: true,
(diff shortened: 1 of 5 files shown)


SebastianRomejko (approved):

src/swagger/client/functional-testing-tools.ts
```diff
@@ -90,6 +91,7 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
     summary:
       "Cancels an ongoing test suite execution in your Swagger Functional Testing workspace. " +
       "Use this tool when you need to stop a long-running or accidentally triggered suite run. " +
+      "The canceled execution is preserved in run history with status `canceled`. " +
```
bartoszkosicki13, line 94: Small suggestion from me: cancelled (two "l"s) instead of cancel — that's what the API returns and what's used elsewhere in the code/tests (functional-testing-api.ts:315).

src/swagger/client/functional-testing-tools.ts
```diff
@@ -146,8 +148,8 @@ export const FUNCTIONAL_TESTING_TOOLS: SwaggerToolParams[] = [
     toolset: "Functional Testing",
     summary:
       "Get the status of a Swagger Functional Testing suite execution. " +
-      "Returns the overall status (pending, canceled, passed or failed), whether the run is finished, and a per-test breakdown with pass/fail. " +
-      "Use this to poll for the outcome of a suite run triggered by `swagger_run_suite`. " +
+      "Returns the overall status (`pending`, `canceled`, `passed`, or `failed`), whether the run is finished, and a per-test breakdown with pass/fail, runtime, and step count. " +
```
bartoszkosicki13: The description says this returns "pass/fail, runtime, and step count" — but per the actual API response shape, each test run also includes `failedStepIndex` (a map of step index -> the actual failure error message, e.g. "Element not found"), not just a count. Worth updating the description to mention this — otherwise an LLM won't know this failure detail is already here and may make an unnecessary extra call to `swagger_get_test_execution_history` just to find out why a test failed.

bartoszkosicki13:
**\`src/swagger/client/functional-testing-types.ts:9\`** (\`testId\` in \`RunFunctionalTestingTestParamsSchema\`, same issue in \`GetFunctionalTestHistoryParamsSchema\`):

list_tests/create_test return id as a number, and the backend consistently models and routes test IDs as a numeric Long (in TestsService.scala — both the run-test and get-test-history routes parse testId via LongVar). This schema, however, requires testId as a plain string. Since the actual contract is numeric, it'd be more accurate to type this as z.coerce.number() (or z.number()) instead of z.string() — non-breaking, since a numeric string still coerces correctly, and it now also accepts the number as returned by list_tests/create_test directly.

---

**\`src/swagger/client/functional-testing-tools.ts:36\`** (\`Create Test\` description, not touched by this PR):

Suggestion: since tests are often built from an OpenAPI/Swagger specification, it'd help to make the mapping explicit, e.g.: "When building tests from a specification, map its \`servers\` entry to \`baseUrl\` and each endpoint's path+method to a step." Right now the description explains the field format but not where those values come from when starting from a spec.

src/swagger/client/functional-testing-tools.ts
KacperSB: It does not return that.
API status enum (`SuiteExecutionStatus.java`) returns "canceled" (one 'l'). If you took this from error message string in the MCP code, it has a typo there - which i will fix.

src/swagger/client/functional-testing-tools.ts
KacperSB: okay, updated

KacperSB:
I will also check for any inaccuracies in both descriptions against aqa, forgot to do that in the first place

ernest-bednarczyk (approved):

PR #706, opened 2026-09-08
[Zephyr] [TM4J-19197] fix response structure schema test cases, cycles and  executions

kordian-smartbear:
## Goal

Fix error of schema mismatch. Current array of objects doesn't match object structure.

## Changeset

Schema was overwritten for functions returning test cases, cycles and executions

## Testing

Unit tests were adjusted. Claude code with local mcp was used to test on the user side.


vmrvictor (approved):

adrienave (approved):

src/zephyr/tool/issue-link/get-test-executions.ts
```diff
@@ -6,6 +7,10 @@ import {
   GetIssueLinkTestExecutions200Response as GetIssueLinkTestExecutionsResponse,
 } from "../../common/rest-api-schemas";
 
+export const UpdatedGetIssueLinkTestExecutionsResponse = zod.strictObject({
+  testExecutions: zod.array(GetIssueLinkTestExecutionsResponse),
+});
+
```
assisblf, line 13: Instead of this, how about we just change `structuredContent: { testExecutions: response },` to `structuredContent: response,`?

With this, we wouldn't need to maintain this, right? Or is this not possible?

btw, This would apply to other cases too

kordian-smartbear: Hi Assis, thanks for checking. Yest this looks like not possible: https://github.com/SmartBear/smartbear-mcp/blob/4706ec1a1c804c63800335025799886df272a6e3/src/zephyr/tool/issue-link/get-test-cases.ts#L41

assisblf (approved):

